PR #5, opened 2016-06-07
Added Arch Linux compile instructions

NicoHood:
@raoulh I can compile moolticuted with those instructions on my raspi3 and start the app. However I still have no access to the device. I tried to run chromium with the official app that works. I used the special archlinux udev rule that is suggested in the official textfile. So there must be something still wrong in the udev rule I guess. Do you have any idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,20 @@ It is completely cross platform, and runs on Linux (using libusb), OS X (native 
 
 ##### Linux
  - Requires libusb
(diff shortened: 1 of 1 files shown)


NicoHood:
Okay I tested again and it seems to work. There might have been some problems on ALARM, but x64 Arch does work now. PR is ready to merge.

raoulh:
thanks

PR #29, opened 2017-02-11
Revert "<udev: Add a udev rules that should benefit for systemd based system>"

bobsaintcool:
This reverts commit 10b1a0b57181302ce42eeabad12399975a12e84f.

Udev rules moved into a dedicated repository https://github.com/bobsaintcool/mooltipass-udev

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Packages are build and available here: https://calaos.fr/mooltipass/
 
 ##### Linux
  - Requires libusb
- - Requires a [udev rule for libusb](https://www.themooltipass.com/udev_rule.txt)
+ - Requires a [udev rule for libusb](https://github.com/bobsaintcool/mooltipass-udev)
 
 ##### Ubuntu 16.04
 ```bash
--- a/daemon.pro
+++ b/daemon.pro
@@ -124,9 +124,4 @@ unix {
     systemd_user.path = $$PREFIX/lib/systemd/user/
     systemd_user.files += $$PWD/systemd/moolticuted.service
     INSTALLS += systemd_user
-
-    # udev rules
-    udev_rules.path = $$PREFIX/lib/udev/rules.d/
-    udev_rules.files += $$PWD/udev/69-mooltipass.rules
-    INSTALLS += udev_rules
 }
--- a/udev/69-mooltipass.rules
+++ b/udev/69-mooltipass.rules
@@ -1,9 +1,0 @@
-# udev rules for allowing console user(s) and libusb access to Mooltipass Mini devices
-ACTION!="add|change", GOTO="mooltipass_end"
-
-# console user
-KERNEL=="hidraw*", SUBSYSTEM=="hidraw", ATTRS{idVendor}=="16d0", ATTRS{idProduct}=="09a0", MODE="0660", TAG+="uaccess"
-# libusb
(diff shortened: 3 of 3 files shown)


NicoHood:
Why dont we add the udev rule to the moolticuted deamon which is the base package for mooltipass then? If i am not wrong the ssh-agent also relies on the deamon, so one can add this as dependency.

A separate repository for the udev rule does not make sense to me. Even if a project needs his own udev rule it could simple overwrite the other. It would be doubled, but that'd be okay if i am not wrong.

This also means that the PKGBUILD should build a split package for moolticuted and moolticute.

bobsaintcool:
> Why dont we add the udev rule to the moolticuted deamon which is the base package for mooltipass then?

Why you don't participate on that point on the AUR mailing list ?

> If i am not wrong the ssh-agent also relies on the deamon, so one can add this as dependency.

Yes, moolticute_ssh-agent relies on moolticuted, but does not requires udev (the agent does care about the device) however the daemon requires the udev rules to works.

> A separate repository for the udev rule does not make sense to me. Even if a project needs his own udev rule it could simple overwrite the other. It would be doubled, but that'd be okay if i am not wrong.

If one does this, there could be a conflict due to the same file being own by two packages, or the same file with different name because different package has been installed with their own rules, it will works but I don't like the game of finding which rules apply

> This also means that the PKGBUILD should build a split package for moolticuted and moolticute.

When you means moolticute you talk about the GUI application ? If yes, I am considering this as an good option that I will probably write soon.

NicoHood:
> Why you don't participate on that point on the AUR mailing list ?

I offered you my help via the mailing list. Just hit me on IRC to chat

> Yes, moolticute_ssh-agent relies on moolticuted, but does not requires udev (the agent does care about the device) however the daemon requires the udev rules to works.

Which other package requires the udev rule then? I think moolticuted should ship the udev rule, as it is the official mooltipass app to access the device soon. Other packages can depend on that or just create another udev rule with the same content but different name. Not optimal, but we would get rid of the extra package which I really would not prefer.

bobsaintcool:
>> Why you don't participate on that point on the AUR mailing list ?

> I offered you my help via the mailing list. Just hit me on IRC to chat

The problem I have with that, is that on the exact same thread you offer you help I discuss the udev topics and I get no answer.

>> Yes, moolticute_ssh-agent relies on moolticuted, but does not requires udev (the agent does care about the device) however the daemon requires the udev rules to works.

>Which other package requires the udev rule then? I think moolticuted should ship the udev rule, as it is the official mooltipass app to access the device soon. Other packages can depend on that or just create another udev rule with the same content but different name. Not optimal, but we would get rid of the extra package which I really would not prefer.

Mooltipy requires it too, even if its not the official application backed by the community its still a close project that provide a different, but interesting approach.

I am not really convinced by your proposals:
* If the udev rule is included into moolticute(d) and tool that need this rule too must depend on moolticute: So you decide to turn my proposal of having both tools have one dependencies by require one tool to fully depend on other package because both tools have one file in common ?
* Use different filename for other package: This require that every packager, of every application on every distribution needs to known which name has already been taken in order to remove the duplicate file on filesystem problem. Also, at some point (even if its not likely the case here) one change must be uploaded into different project.

Finally I thing I don't understand the problem of adding one dependencies (on every package) that make you offer such solution ?

PR #30, opened 2017-02-11
Dev/systemd: Move the systemd service file systemd wide

bobsaintcool:
Only one instance of moolticute could be run at a time. It makes more
sense to move the systemd service file systemd wide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Packages are build and available here: https://calaos.fr/mooltipass/
 
 ##### Linux
  - Requires libusb
- - Requires a [udev rule for libusb](https://www.themooltipass.com/udev_rule.txt)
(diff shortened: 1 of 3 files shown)


NicoHood:
I cannot find the moolticute PKGBUILD anywhere on aur. Am I missing something? I'd like to test it, would you like to upload it and add me as co-maintainer? If its popular enough I can push it to community.

bobsaintcool:
For now PKGBUILDs are on my personal repo[0][1], they probably goes on AUR today for the git version. In other hand, for the released based PKGBUILD we still needs some bug correction.

https://git.bourgeois.eu/aur_moolticute.git/
[1] https://git.bourgeois.eu/aur_moolticute_git.git/

PR #32, opened 2017-02-13
Work In Progress: Correcting Corrupted Data Flash

limpkin:


src/MPDevice.cpp
```diff
@@ -1773,7 +1791,7 @@ bool MPDevice::addOrphanParentToDB(MPNode *parentNodePt)
     else
     {
         /* It is important to note that this function is called with a valid linked chain (due to tagcheck) */
-        for (auto &i: loginNodes)
+        for (auto &i: list)
```
raoulh: Please use a different name for "list" here. list is often refered as std::list. also if one does something like `use namespace std;` there will be a name clash

raoulh: Please use a different name for "list" here. list is often refered as std::list. also if one does something like `use namespace std;` there will be a name clash

limpkin: noted :)

PR #43, opened 2017-07-04
Pull request for ongoing development

limpkin:


src/MPDevice.cpp
```diff
@@ -1994,7 +1994,7 @@ bool MPDevice::addChildToDB(MPNode* parentNodePt, MPNode* childNodePt)
         else
         {
             /* If the login name is supposed to be before the one we're trying to insert */
-            if (tempNextChildNodePt->getLogin() > childNodePt->getLogin())
+            if (tempNextChildNodePt->getLogin().compare(childNodePt->getLogin()) > 0)
```
raoulh: Maybe better using QCollator here for comparing strings?
http://doc.qt.io/qt-5/qcollator.html
https://wiki.qt.io/QCollator

limpkin: I'm still not sure about that... the last time I tried it the result ended up being incorrect because of locales :/

raoulh: the locale can be set and forced to the qcollator if it's better

limpkin: I'd rather keep it simple & stupid :)

raoulh: The problem here is not to keep it simple, but make it correct. And the Qt doc explicitly says that QString::compare is not what a human would expect:

> Case sensitive comparison is based exclusively on the numeric Unicode values of the characters and is very fast, but is not what a human would expect. Consider sorting user-visible strings with localeAwareCompare().

http://doc.qt.io/qt-5/qstring.html#compare

A good solution is to use QCollator that allows more natural sorting of string containing numbers like:
blabla_1
blabla_2
blabla_10

Using compare or standard the list would be sorted as:
blabla_1
blabla_10
blabla_2

Which is wrong for us humans.

limpkin: Well the thing is that the mooltipass is doing exactly like compare(), so if we use qcollator the sorting will be completely different :/

raoulh: the MP is doing the same as the Qt QString::compare() ? (it's not the C version from the std lib)

limpkin: MP is following the ASCII code numbering yes.

limpkin: here's the code line: https://github.com/limpkin/mooltipass/blob/master/source_code/src/NODEMGMT/node_mgmt.c#L1195

PR #62, opened 2017-08-04
Files cache

azubieta:
Implementation of the file cache, still missing the "getStoredFiles" function which should retrieve the names file of the files stored in the card.
Also the "updateFilesCache" slot should be called when entering and leaving MMM.


limpkin:
closed because #63 implements missing features & solves bugs.
Do not forget to update your repo.

PR #67, opened 2017-08-15
Combine device information labels at one widget

medyakovvit:
It allows to hide/show all labels with device information through changing visibility of widget. It also fix "%1" in device information labels at "About" page.

When you start gui application without an device connected, you can see this on "About" page:
aboutpage


raoulh:
Thanks!

limpkin:
Thanks a lot!

PR #82, opened 2017-08-28
Change files tab and the ssh keys tabs visibility options

azubieta:
Set the files tab visible by default and keep the ssh keys tabs visible on demand or allways behaviour


limpkin:
@raoulh : ctrl-shift-f2 is the shortcut you'll need. I'm merging as this is something we've already discussed and agreed upon.

PR #84, opened 2017-08-29
OSX .dmg GitHub upload

tomasz-janiczek:


limpkin:
@raoulh : this PR is mainly meant as a bug fix for the previous PR that @tomasz-janiczek which was not working due a tag related problem. Could you therefore review, merge and create a new release as @azubieta also recently added needed bug fixes and features to Moolticute? I'll then make a call for beta testers.

raoulh:
Hi,
Build is failing. I cannot merge that.
Also please squash your commits to be easier to follow.

.travis.yml
```diff
@@ -8,15 +8,13 @@ os:
...
-  - "git fetch --unshallow"
-  - ". scripts/ci/$TRAVIS_OS_NAME/before_install.sh"
-install: ". scripts/ci/$TRAVIS_OS_NAME/install.sh"
+  - make -f Makefile.travis before_install
+install:
+  - make -f Makefile.travis install
 script:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make debian DEBUG=1; fi
-  - ". scripts/ci/$TRAVIS_OS_NAME/script.sh"
+  - make -f Makefile.travis script
```
raoulh: No, I'm against that change. It does not help in any way to add a new makefile that calls the exact same script. It's just another indirection and source of problems/errors. It just make the builds even more complicated.

Please do not change the actual way of build the project for windows and osx. Docker for ok for the deb, but it should be added in parallel of the actual build, the same way it was added to the .travis.yml
Think that it should be easy to understand when someone needs to put its hand inside the biuld system again that it should remain easy to understand and has the same structure for all build

docker/Dockerfile.windows
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:16.04
```
raoulh: What is that file used for?
The windows build is already done in the existing script.
Please don't rewrite thing that works your way. it's not helpful at all.

Makefile
```diff
@@ -49,18 +52,34 @@ build: docker_prepare
...
+	docker build -t $(PROJECT_NAME).windows -f docker/Dockerfile.windows .
 
 # GitHub
 github_upload:
-	$(DOCKER_EXEC) ". /usr/local/bin/tools.sh && create_release_and_upload_asset $(TRAVIS_TAG) $(DEB_NAME) $(DEB_MIME)"
+ifneq ($(BUILD_TAG),)
+ifeq ($(TRAVIS_OS_NAME),linux)
+	$(DOCKER_EXEC) "source docker/tools.sh && \
+	PROJECT_NAME=$(PROJECT_NAME) TRAVIS_OS_NAME=$(TRAVIS_OS_NAME) \
+	create_github_release $(BUILD_TAG)"
```
raoulh: This needs to go in the `scripts/ci/linux/after_success.sh` after the other upload code is done. Not in another makefile started differently

Makefile
```diff
@@ -49,18 +52,34 @@ build: docker_prepare
...
+else
+	brew install jq
+	source docker/tools.sh && \
+	osx_setup_fake_home /tmp && \
+	export PATH=$(PATH):$(shell pwd)/scripts/lib && \
+	HOME=/tmp PROJECT_NAME=$(PROJECT_NAME) TRAVIS_OS_NAME=$(TRAVIS_OS_NAME) create_github_release $(BUILD_TAG) && \
+	osx_clean_fake_home /tmp
```
raoulh: Same here for macos.
And the installation of tools deps needs to be done in `scripts/ci/osx/before_install.sh`

Makefile.travis
```diff
@@ -0,0 +1,27 @@
+SHELL := /bin/bash
```
raoulh: No need for this file at all.

docker/tools.sh
```diff
@@ -14,7 +14,8 @@
 #
```
raoulh: Also merge this file in `scripts/ci/funcs.sh`
There is no need to have separated files for functions used by multiple builds.

.travis.yml
tomasz-janiczek: Hi @raoulh 
Your code your rules :) But let's please agree for the (potential) future that you'll provide me with the guidelines and requirements before I start the work, not after I've basically finished it. No need to waste time and effort on something that isn't working as you want it to, right?
I'll be looking into your remarks today afternoon/evening.
Thanks.

limpkin:
that would actually be a mistake on my side, as I should have consulted
with @raoulh before telling you to go on.

.travis.yml
tomasz-janiczek: About the makefile - valid point, agreed.
Wrote it assuming the Travis-related code will grow significantly in the future so it would be nice to keep the .travis.yml file nice and clear by adding another layer of abstraction and maybe creating a makefile on a 'per OS' basis (or something in that direction).

Either way - I'll change it.

docker/Dockerfile.windows
tomasz-janiczek: That's exactly the cause for this ;) Sure, not a problem.
The idea was to eventually create two pre-defined Docker images with all the needed libs and tools pre-installed (two images = lower size = faster download = quicker build times/effects). Dropping the idea.

Makefile
tomasz-janiczek: Sure, if we want to stick to the original scripts and have it all in one place.
I've intentionally kept them as they were (the scripts), thus the makefile approach. Additionally the makefiles make it easier to run just parts of the different stages/events of the build (helps with testing etc.). Dropping it also.

Makefile
tomasz-janiczek: Valid point.

Makefile.travis
tomasz-janiczek: Ok, removing.

docker/tools.sh
tomasz-janiczek: Again - didn't want to modify your scripts directly.
Ok.

tomasz-janiczek:
Closing this PR and will open a new one, for the refactored code.

PR #86, opened 2017-08-30
Improve progress feedback

azubieta:


limpkin:
@raoulh : do not merge, WIP.

azubieta:
About the commit 576f447  at ./src/MPDevice.cpp:6421 we should get the real file size instead of the value I'm using.
Didn't dig forward because i was told that someone else will do that part.

src/MPDevice.cpp
```diff
@@ -6402,12 +6410,18 @@ void MPDevice::getStoredFiles(std::function<void (bool, QStringList)> cb)
...
 
         /* Get file names */
         for (auto &i: dataNodes)
         {
-            list.append(i->getService());
+            QVariantMap item;
+            item.insert("name", i->getService());
+
+            // FIXME: This isn't the actual file size we should get the real one.
+            item.insert("size", i->getDataNodeData().length());
```
azubieta: Heads up here, this isn't the real file size. It should be replaced with the real one.
I was told that someone else will do it so I didn't go forward.

src/MPDevice.cpp
```diff
@@ -4203,7 +4211,8 @@ void MPDevice::setDataNode(const QString &service, const QByteArray &nodeData,
         cb(true, QString());
 
         // update file cache
-        addFileToCache(service);
+        // FIXME: Set the real file size
+        addFileToCache(service, 0);
```
azubieta: Again, we should put the real file size here.

src/MPDevice.cpp
limpkin: we'll talk about that.

limpkin:
@raoulh : this has been **thoroughly** tested and is ready to merge, **before** our beta testers reelase. I'm guessing you'll also need to make sure that your CLI si compatible. 
In short: we simply added a text to the cbprogress callback for cred/file MMM, import, export, integrity check.
Moreover, the first arg of cbprogress callback can now be -1 (always moving progress bar)

raoulh:
Thanks

PR #87, opened 2017-09-01
Debian package & GitHub upload

tomasz-janiczek:


scripts/ci/linux/after_success.sh
```diff
@@ -59,4 +59,24 @@ EOF
...
+$DOCKER_EXEC "dpkg-buildpackage -b -us -uc && mkdir -p build-linux/deb && cp ../*.deb build-linux/deb"
+
+# GitHub release
+$DOCKER_EXEC \
+    "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG} PROJECT_NAME=${PROJECT_NAME} TRAVIS_OS_NAME=${TRAVIS_OS_NAME} \
+    DEB_VERSION=${DEB_VERSION} && \
+    source /usr/local/bin/tools.sh && \
+    create_github_release ${BUILD_TAG}"
+
+#popd
```
raoulh: You could also add the upload to github of the windows build here: `build/$FILENAME.exe`

limpkin:
I may be mistaken, but with this PR the .exe also gets uploaded:
https://github.com/limpkin/moolticute/releases

src/version.h
```diff
@@ -1,4 +0,0 @@
-#ifndef VERSION_H
```
raoulh: Why this file is removed? it's needed for development

tomasz-janiczek: My bad, commited this by mistake.

scripts/ci/linux/after_success.sh
tomasz-janiczek: Why not. I'll move it.

tomasz-janiczek: @limpkin is right - I forgot I already implemented the .exe upload :) It's there.

scripts/ci/funcs.sh
```diff
@@ -36,3 +62,182 @@ function upload_file()
...
+    fi
+
+    >&2 echo -e "Creating GitHub release (tag: $VERSION)"
+
+    if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+	    create_release_and_upload_asset $VERSION $DEB_FILE
+	    create_release_and_upload_asset $VERSION $EXE_FILE
+    elif [ "$TRAVIS_OS_NAME" == "osx" ]; then
+	    create_release_and_upload_asset $VERSION $DMG_FILE
+    fi
```
raoulh: Please avoid that indirection. This is a common script.
the if [ "$TRAVIS_OS_NAME" == "linux" ] .. is already done earlier. To keep the code clear this uploads should be done is the corresponding after_install.sh scripts next to the other uploads.

scripts/ci/linux/after_success.sh
```diff
@@ -59,4 +59,24 @@ EOF
...
+
+$DOCKER_EXEC "cp -f README.md debian/README"
+$DOCKER_EXEC "dpkg-buildpackage -b -us -uc && mkdir -p build-linux/deb && cp ../*.deb build-linux/deb"
+
+# GitHub release
+$DOCKER_EXEC \
+    "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG} PROJECT_NAME=${PROJECT_NAME} TRAVIS_OS_NAME=${TRAVIS_OS_NAME} \
+    DEB_VERSION=${DEB_VERSION} && \
+    source /usr/local/bin/tools.sh && \
+    create_github_release ${BUILD_TAG}"
```
raoulh: Please avois using `&&` for no reason. `set -e` is already used to stop the script.
it's also not good for maintaining this code, as I was thinking it was only for the docker build, but you do the upload also here....

PR #103, opened 2017-09-16
Using the AppGui awesome instance instead of creating a new one.

### LINKED ISSUE #90: fontawesome icons are wrong in release
In released build, icons from fontawesome are wrong or not displayed

image

Password linedit should have an Eye + a tool icon, instead there is an arrow on windows. On macos no icons are displayed but two squares.

azubieta:
Fix #90


raoulh:
Good catch!

PR #104, opened 2017-09-18
Adding a thread for the libusb events loop

azubieta:
By adding a new thread for handling the libusb events loop when a device is connected the speed is increased to 27 kb/s while performing integrity check.


limpkin:
@raoulh : could you look attentively to this PR? this is quite major and needs to be double checked.

raoulh:
Thanks, great work!

PR #106, opened 2017-09-20
Use icons path compliant to freedesktop specs

nemunaire:
This pull request aims at fixing QA issue required to be embedded into distributions (see last comment of https://github.com/gentoo/gentoo/pull/5520).

[According to freedesktop.org](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html), icons should be placed in theme, not at the root of `/usr/share/icons`. `/usr/share/pixmaps` could be used like that instead, but this is not recommanded.

Depending on your environment, you could have to refresh the icon cache by calling something like `gtk-update-icon-cache /usr/share/icons/hicolor`. This is generally done by distribution package install script, when needed (like calls to `udevadm control --reload-rules`).


raoulh:
Thanks

PR #108, opened 2017-09-20
Add hardening settings to systemd service

nemunaire:
This pull request adds a lot of hardening features to moolticuted systemd service.

The most important one is the execution as normal user instead of root. For this use, it requires good permissions defined on USB device (through given udev rule).

Others settings indicate unused features, to enforce their non-use in case of vulnerability.

### DIFF
--- a/systemd/moolticuted.service
+++ b/systemd/moolticuted.service
@@ -2,10 +2,30 @@
 Description=Moolticute daemon
 
 [Service]
+Type=simple
 ExecStart=/usr/bin/moolticuted
 KillMode=process
 Restart=always
-Type=simple
+
+# Uncomment those fields and edit them to match your settings
+#User=nobody
+#Group=plugdev
+
+# Hardening for all
+CapabilityBoundingSet=
+RuntimeDirectory=moolticuted
+RuntimeDirectoryMode=750
+PrivateTmp=yes
+RemoveIPC=true
+ProtectSystem=strict
+ProtectHome=read-only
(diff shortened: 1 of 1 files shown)


nemunaire:
As pointed out by @raoulh, `usb` group is not commonly used (eg. `plugdev` on debian).

I made following changes to my PR, as a trade off between security and out-of-the-box use on every system.

```diff
--- a/systemd/moolticuted.service
+++ b/systemd/moolticuted.service
@@ -6,9 +6,13 @@ Type=simple
 ExecStart=/usr/bin/moolticuted
 KillMode=process
 Restart=always
-User=nobody
-Group=usb
-# Hardening
+
+# Uncomment those fields and edit them to match your settings
+#User=nobody
+#Group=plugdev
+
+# Hardening for all
+CapabilityBoundingSet=
 RuntimeDirectory=moolticuted
 RuntimeDirectoryMode=750
 PrivateTmp=yes
```

With this new `.service`, the daemon can access the mooltipass even without udev rules (because the user `root` has RW access to devices), but thanks to `CapabilityBoundingSet=`, this is no more a super-user under `moolticute` execution session.
The only advantage to uncomment `User` and `Group` lines is because even with less capabilities, many files are owned by `root` and still remain accessible. Executing as `nobody` prevents access to them, and is mostly limited to `plugdev` group files.

I think we should keep this as simple and uncomment `User` and `Group` only when we generate packages targeted to a particular distribution.

raoulh:
Thanks

PR #120, opened 2017-10-10
Refreshing language selection combo box on language changed.

azubieta:


limpkin:
@raoulh can be merged if you think so.

PR #123, opened 2017-10-11
Add note about what C++ compiler to use on Windows

NoahAndrews:


NoahAndrews:
If you want to know, here's a gist of the output I currently get when compiling using the Visual C++ 2017 compiler: https://gist.github.com/NoahAndrews/5608c0119ec6c36d337ceba7f1b72a0d#file-msvc2017-log-L121

PR #127, opened 2017-10-13
Encryption of exported db files.

azubieta:


limpkin:
tested and approved!
@azubieta will make a new PR to replace the text about encryption

PR #139, opened 2017-10-28
Implement updates using GitHub releases

azubieta:


limpkin:
@raoulh : tested on WIN on my side :)

NicoHood:
And where is the GPG signature check of the update file? How do you protect on MITM except of blindly relying on https and the integrity of the github releaese files?

limpkin:
@NicoHood : win has digital signatures enabled, macOS is about to, linux doesn't auto download the files.

raoulh:
I agree with @NicoHood, there is no GPG check right now when after downloading the binaries. Is should be done.

@limpkin the digital signature of binaries is not for the same purpose.

limpkin:
@raoulh could you expand why it is not for the same purpose? having a signed installer guarantees by definition that it is the correct one no?

raoulh:
No because we are not checking if the certificate use for the signing is ours. It's only for the OS that will display a warning if the signature is wrong.
Also adding code for checking this signature would be a pain. It's heavily OS dependent and would require a lot of work to develop that on each platform.

The basic problem right now is MITM. If we download a binary which is signed by someone else with a valid certificate, the installation procedure will go through without any warning and it will not be OUR package.

The easiest way of working around that issue is simple GPG signature of our binaries.

limpkin:
I see! 
For mooltiapp, the update procedure would be transparent (not requiring the user to click anything) as long as the signature was correct. I'm guessing there's no similar module for qt?

PR #141, opened 2017-10-31
Integrate all thirdparty software

azubieta:
WIP in order to get a copy of all third party software used by moolticute.
I'll need this to be merged in order to get the right links.


limpkin:
@azubieta : at the moment you may still allow download from @raoulh 's server. Could you therefore update this PR to only fetch the repositories from our forks?

raoulh:
Not valid anymore, closing it

PR #146, opened 2017-11-09
Pack as appimage testing

azubieta:


azubieta:
My bad this should be done at my own github repo

PR #148, opened 2017-11-10
Appimage wrapper

azubieta:
Add a wrapper script to install Udev rules if they don't exists.
Also allows uninstall, check rules, and run as daemon.

### DIFF
--- a/data/moolticute.sh
+++ b/data/moolticute.sh
@@ -1,15 +1,20 @@
 #!/bin/bash
-
+##
+# Wrapper for the moolticute AppImage
+##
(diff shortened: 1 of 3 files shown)


data/moolticute.sh
```diff
@@ -0,0 +1,112 @@
...
+
+Example:
+Run Moolticute normaly
+./$me
+
+Run only the moolticuted daemon
+./$me --daemon
+";
+
+SUDO_MSG="Allow Mooltipas Udev rules to be applied in your system?"
```
raoulh: Mooltipass typo

PR #153, opened 2017-11-16
Db backups tracker

azubieta:
"Backend" of the backups tracking functionality.


src/DbBackupsTracker.cpp
```diff
@@ -0,0 +1,217 @@
...
+    return content;
+}
+
+int DbBackupsTracker::extractCredentialsDbChangeNumber(const QString& content) const
+{
+    int cn = 0;
+    QJsonDocument d = QJsonDocument::fromJson(content.toLocal8Bit());
+    if (d.isObject()) {
+        QJsonObject root = d.object();
+        if (root.contains("credentialsDbChangeNumber"))
```
limpkin: are you sure that works for both encrypted and non encrypted versions of the export file?

limpkin:
Nice code! However it may be nice in the future to comment your code :/

limpkin:
@azubieta is it ready to merge or would you like to implement the calls to this new file in mpdevice before we merge?

azubieta:
I guess that the integration with the MP device should be done at the time
of making the UI.

I would like to give it another review today. But there should not be major
changes in the class interface.

limpkin:
@azubieta would you be able to merge your work with @medyakovvit other standing PR which already contains the UI part?

src/DbBackupsTracker.cpp
azubieta: Fixed in d81d76c

azubieta:
This is almost done, there is only missing the integration of the previous implementation message dialogs. Please test it.

Also there is an issue about QFileSystemWatcher. When the file is remove it stops tracking it. Our options:

1. leave it, this is a best effort functionality.
1. track also the directory containing the file (if the directory also get removed the watch stops)
1. put a timed loop that reset the tracking every x seconds.

limpkin:
that may be a stupid question... but how can we test it if there are no
dialogs?

azubieta:
There are dialogs but not the fancy ones. All the functionality is
implemented.

limpkin:
Oh great!

Could you check your implementation against these tests?
https://github.com/mooltipass/moolticute/pull/85#issuecomment-339413852

Thanks

azubieta:
Checked, all tests passed!

limpkin:
awesome! will test it now.

limpkin:
- I import a file X
- I specify file X to be monitored
in gui console:
shell\comdlg32\fileopensave.cpp(14267)\comdlg32.dll!76267C7D: (caller: 762968FC) ReturnHr(9) tid(2498) 80004005 Unspecified error
    CallContext:[\PickerModalLoop] 
- I import file X again
- monitored file field is empty

we're really not up to a great start :/

Please also pull from master. To increment the device credential change number during your test, simply add a credential.

azubieta:
I track down the windows issue, it happen that the default constructor of the qsettings that as the doc says should work. It doesn't so I do it using the long one.

azubieta:
With 4cca87a the previous Ui gets totally integrated with my code.

limpkin:
Given 2 backup files: file X for database at moment T, file Y for database at moment T+1 (with credchangenumber incremented)

Scenario 1:
- X imported to device
- setting up monitored file to be X
- adding new credential to device through the credentials tab
- **monitored file disappears**

Scenario 2:
- X imported to device
- setting up monitored file to be X
- importing file Y
- **monitored file disappears**

seems you didn't track down the issue.

src/DbBackupsTracker.cpp
```diff
@@ -0,0 +1,317 @@
...
+void DbBackupsTracker::refreshTracking()
+{
+    if (tracks.contains(cardId))
+        track(tracks.value(cardId));
+}
+
+void DbBackupsTracker::saveTracks()
+{
+    QSettings s(QSettings::IniFormat, QSettings::UserScope, qApp->organizationName(), qApp->applicationName());
+    s.beginGroup("BackupsTracks");
```
azubieta: Debug file is being created
qDebug() << s.fileName() ;

limpkin:
ok, that works beautifully! congratulations to you both @azubieta & @medyakovvit 

@azubieta : one last thing, the change number is only incremented **once** per session. This means that we don't get the prompt to update the monitored file if we add 2 credentials. Could you therefore trigger the notification every time a new credential is added here? 
https://github.com/mooltipass/moolticute/blob/master/src/MPDevice.cpp#L4076

@raoulh : tested on my side. Could you review and merge?

limpkin:
@raoulh that concludes this PR. Ready for your final review and merge.

limpkin:
@azubieta : could you set the export file format to be **the exact same** as the current format of the monitored file?

limpkin:
so happy we got there! just a quick release and i'll make a public announcement :)

PR #155, opened 2017-11-27
Hide backup import and export messages 

azubieta:
Hide import and export messages when:
 - the device is disconnected, 
 - the card is removed or
 - the current db backup file being track is changed.


limpkin:
@raoulh looks good to me (tested), could you review and merge?

PR #158, opened 2017-12-08
db backups traking  only in firmwares equal or greater than 1.2

azubieta:
Backups tracking controls are hidden for devices with firmwares lower that "1.2"
An alphanumeric comparison is performed as the version number is an string.

Please check


limpkin:
Can be merged then :)

PR #159, opened 2017-12-14
Notify import on change numbers wrap over.

### LINKED ISSUE #156: Handle credentialchangenumber wrap-over
@azubieta contact me for more details

azubieta:
fix #156


limpkin:
thanks! did you test it?

src/DbBackupsTracker.cpp
```diff
@@ -162,8 +162,7 @@ bool DbBackupsTracker::isBackupRequired() const
         int backupCCN = tryGetCredentialsDbBackupChangeNumber();
```
limpkin: does this fail when the device has lower than v1.2 firmware?
is this code guaranteed to run **after** the change numbers are loaded from the device when a device is disconnected and another is plugged?

src/DbBackupsTracker.cpp
```diff
@@ -271,17 +270,39 @@ void DbBackupsTracker::setDataDbChangeNumber(int dataDbChangeNumber)
...
+
+    // *and* the difference between both numbers is less than 0x60
+    result = result && ( abs(backupCCN - credentialsDbChangeNumber) < 0x60);
+
+     /*
+     * - the monitored file change number is < 0x20 and the
+     * change number on the device is > 0xE0.
+     */
+    result = result ||
+            ((backupCCN < 0x20) && (credentialsDbChangeNumber > 0xE0));
```
limpkin: please also handle datachange number.

src/DbBackupsTracker.cpp
```diff
@@ -162,8 +162,7 @@ bool DbBackupsTracker::isBackupRequired() const
         int backupCCN = tryGetCredentialsDbBackupChangeNumber();
         int backupDCN = tryGetDataDbBackupChangeNumber();
 
-        return (backupCCN < credentialsDbChangeNumber || backupDCN < dataDbChangeNumber);
-
+        return isDbBackupChangeNumberLower(backupCCN, backupDCN);
```
limpkin: this will fail when the monitored file has 0xFF and the number on device wraps over to 0x00.
why not !isDbBackupChangeNumberGreater?

src/DbBackupsTracker.cpp
```diff
@@ -270,26 +270,34 @@ void DbBackupsTracker::setDataDbChangeNumber(int dataDbChangeNumber)
     emit dataDbChangeNumberChanged(this->dataDbChangeNumber);
 }
 
-bool DbBackupsTracker::isDbBackupChangeNumberGreater(int backupCCN, int backupDCN) const
+bool DbBackupsTracker::greaterThanWithWrapOver(int a, int b, int limit, int range) const
 {
-    // - the monitored file change number is higher than the one the device
-    bool result = (backupCCN > credentialsDbChangeNumber || backupDCN > dataDbChangeNumber);
+    bool res = (a > b) && ( (a - b) < range);
+    res = res | ((limit - b + a) < range);
```
limpkin: why no (a < b) &&  ((limit - b + a) < range); ?

azubieta: no need of (a < b), (limit -b + a) < range will be false any way, but if you want we could add the check.

src/DbBackupsTracker.cpp
azubieta: I do test it making the daemon to return a version number lower than v1.2 but didn't test it with a real device.

src/DbBackupsTracker.cpp
azubieta: Fixed in 6b9241d

limpkin:
Please do.

azubieta:
Done

PR #161, opened 2017-12-15
Correct minimum firmware version number to show the import dialog.

### LINKED ISSUE #160: PR #158 does not work
@azubieta : I don't know if/how you've tested your code but I'm still being displayed the "Database Backup" for a device that has the "v1" firmware.
Please correct that.

azubieta:
Fix #160
there was an error in the version number that i was using as base to show the import dialog.
The "v" was missing.
Please notice that this is a lexicographical comparison.


limpkin:
:/

Issue #163, opened 2017-12-15
Disable Quick File Access for device that have < v1.2 firmware


limpkin:
Similarly to disabling the Database Backup Monitoring functionality....

limpkin:
for @azubieta

limpkin:
reason: we don't have the change number and therefore can't keep an updated list.

PR #164, opened 2017-12-18
Fix file quick access

### LINKED ISSUE #162: File Quick Access is not working
When you have a file listed in Quick File Access in the file management tab, export is not working.
When you click the icon to save it on disk, a filename is prompted, then "Data Fetch Denied" error.

In log file:
DEBUG: :0 - JSON API recv: {
    "data": {
        "service": ""
    },
    "msg": "get_data_node"
}
WARNING: :0 - context is empty.

The context sent to the daemon is empty, so no prompt.

When the call to the json api fails, do not continue and ask for a filename.

azubieta:
Fix #162 and #163

### DIFF
--- a/src/FilesManagement.cpp
+++ b/src/FilesManagement.cpp
@@ -139,10 +139,21 @@ void FilesManagement::setWsClient(WSClient *c)
     {
         loadFilesCacheModel();
     });
+
+    setFileCacheControlsVisible(wsClient->get_fwVersion() >= "v1.2");
+    connect(wsClient, &WSClient::fwVersionChanged, [=](const QString &version)
+    {
+        setFileCacheControlsVisible(version >= "v1.2");
+    });
     connect(wsClient, &WSClient::wsConnected, [=] ()
     {
         wsClient->sendListFilesCacheRequest();
     });
+}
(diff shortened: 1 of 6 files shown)


src/FilesManagement.cpp
```diff
@@ -139,12 +139,23 @@ void FilesManagement::setWsClient(WSClient *c)
     {
         loadFilesCacheModel();
     });
+
+    setFileCacheControlsVisible(wsClient->get_fwVersion() >= "v1.2");
```
raoulh: This is not good. String compared like that is likely to fail. The mini version contains also `_mini` at the end.

Please implement wsClient->isFw12() like in MPDevice.

See the implementation in the backend:
https://github.com/mooltipass/moolticute/blob/02a087d2d5b16bb47ead47f35af239f71ef6995a/src/MPDevice.cpp#L400
and 
https://github.com/mooltipass/moolticute/blob/e8b7dd5894f80cff0d1c44689454f877214f4c66/src/MPDevice.h#L211

azubieta:
@raoulh firmware comparison was implemented as you requested please check it.

Issue #165, opened 2017-12-21
Automatically lower case service names during add and edition in MMM


limpkin:
As the mooltipass device automatically lowers any service name going his way, we need to do the same when adding or editing service names in MMM

limpkin:
for @azubieta

limpkin:
when to trigger lowercasing operation:
- credentials tab, in MMM and not in MMM: when either "enter" is pressed or "save" is clicked in the "add or update a credential" panel
- in files tab, in MMM and not in MMM: when either "enter" is pressed or "save" is clicked in the "add or update a file" panel

azubieta:
The idea here is to ensure to lower case every service name that is passed to the daemon ?

limpkin:
yes, but do the conversion in the GUI at the moments specified above.

raoulh:
Also all service names must be lowercased in the daemon when they comme from the Json API

raoulh:
Yes, as all service names should be lowercase only, it's better to also enforce lowercase in the api

PR #167, opened 2017-12-22
Transform to lower case all service names before they are sent to the device.

azubieta:
Implements #165

### DIFF
--- a/src/CredentialsManagement.cpp
+++ b/src/CredentialsManagement.cpp
@@ -229,7 +229,7 @@ void CredentialsManagement::on_addCredentialButton_clicked()
 {
     if (wsClient->get_memMgmtMode())
     {
-        m_pCredModel->addCredential(ui->addCredServiceInput->text(),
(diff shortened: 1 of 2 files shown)


limpkin:
@raoulh did you review that?

limpkin:
can we merge that one and write in our todo list to implement it in the daemon as well? :)

limpkin:
ping @raoulh

raoulh:
This PR is wrong, i pushed a correct fix for this problem in master

limpkin:
@raoulh are you sure? 
Consider the following scenario:
- service "foo" is in the device
- in MMM, user wants to add "Foo"

what will then happen?

raoulh:
I already pushed a better fix for all those cases in master

limpkin:
yep, I know that :).
But in your better fix, is the scenario described above correctly handled?

raoulh:
Of course.

limpkin:
awesome :D

Issue #182, opened 2018-01-08
Simple Feature Requests


limpkin:
1) Files tab: when selecting a file, when the service/file name field is empty, the service/file name field gets automatically filled with the file name
2) Files tab: when selecting a file, have the file picking window go to the user main's directory instead of the moolticute installation directory

PR #184, opened 2018-01-12
Fix Issue 182

azubieta:
implements #182


limpkin:
on a professional computer, this PR actually does select the admin account of my session... is that normal?

Issue #178, opened 2018-01-05
Feature Request: Help popups


limpkin:
In device settings:
- Add a text along the lines of "Hint: keep your mouse positioned over an option to get more details"
- then implement some help text for each option

In synchronization tab:
- Next to each title (Export Database to File....) add a "?" icon, clicking on it brings an overlay with an explanation of the feature

azubieta:
Is a tooltip like the one in the picture OK for the synchronization help ? Or you want a larger dialog ?
![screenshot_20180112_113318](https://user-images.githubusercontent.com/1138094/34887508-636cbea4-f78c-11e7-94d6-0fe38bdd342f.png)

azubieta:
@limpkin I'll need some help writing the "help" messages

limpkin:
Put place holders I will put the text after

azubieta:
Placeholders have been set.

PR #185, opened 2018-01-12
Help popups.

azubieta:
Implements #178 

THIS IS A WORK IN PROGRESS DO NOT MERGE

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -42,16 +42,16 @@ void MainWindow::initHelpLabels()
     };
 
     ui->label_exportDBHelp->setPixmap(getFontAwesomeIconPixmap(fa::questioncircle));
-    ui->label_exportDBHelp->setToolTip(tr("Placeholder help hint."));
+    ui->label_exportDBHelp->setToolTip(tr("Export your encrypted database to a file"));
 
     ui->label_importDBHelp->setPixmap(getFontAwesomeIconPixmap(fa::questioncircle));
-    ui->label_importDBHelp->setToolTip(tr("Placeholder help hint."));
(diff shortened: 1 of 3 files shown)


src/MainWindow.cpp
```diff
@@ -42,16 +42,16 @@ void MainWindow::initHelpLabels()
     };
 
     ui->label_exportDBHelp->setPixmap(getFontAwesomeIconPixmap(fa::questioncircle));
-    ui->label_exportDBHelp->setToolTip("Placeholder help hint.\n");
+    ui->label_exportDBHelp->setToolTip(tr("Placeholder help hint."));
```
azubieta: Change the placeholder text for the real help message.

azubieta:
For the Synchornization tab change the placeholder text for the real help messages [here](https://github.com/mooltipass/moolticute/pull/185/commits/901b13c40d67e238efd1bce6e9011fc26ee9739a#diff-1c0b64226a129d35f02dc0049484cd6bR45)

For the device settings tab use the ui editor and modify the tooltip property of every settings widget item.

limpkin:
@raoulh how would you prefer to proceed for that PR? Would you like me to work on @azubieta's branch or can you merge that one and I'll add the texts just after?

limpkin:
@azubieta I don't see any tooltip for the individual settings... is that normal?

azubieta:
For the individual settings you should modify the tooltip field in the widget settings. This is done using the qt creator UI editor.
For the demonstration of how it's done I just filed the first two settings items. The rest still has the same filed.

azubieta:
For the individual settings you should modify the tooltip field in the widget settings. This is done using the qt creator UI editor.
For the demonstration of how it's done I just filed the first two settings items. The rest still has the same filed.

PR #195, opened 2018-01-20
fix for problem with sizeof(MPCmd::Command) #192

sbobykin:
Fix for the issue #192


raoulh:
Thanks

sbobykin:
@raoulh thank you too)

Issue #149, opened 2017-11-13
AppImage not working


raoulh:
On Archlinux:

```
╭─raoul@raoul-laptop ~/Downloads  
╰─➤  ./Moolticute-v0.9.19-beta.glibc2.14-x86_64.AppImage                                                                                                         127 ↵
zenity, kdialog, Xdialog missing. Skipping /tmp/.mount_MooltibpsDfv/usr/bin//moolticute.sh.wrapper.
which: no kdesudo in (/tmp/.mount_MooltibpsDfv/usr/bin/:/tmp/.mount_MooltibpsDfv/usr/sbin/:/tmp/.mount_MooltibpsDfv/usr/games/:/tmp/.mount_MooltibpsDfv/bin/:/tmp/.mount_MooltibpsDfv/sbin/:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)
which: no gksudo in (/tmp/.mount_MooltibpsDfv/usr/bin/:/tmp/.mount_MooltibpsDfv/usr/sbin/:/tmp/.mount_MooltibpsDfv/usr/games/:/tmp/.mount_MooltibpsDfv/bin/:/tmp/.mount_MooltibpsDfv/sbin/:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)
Allow Mooltipass Udev rules to be applied in your system?
[sudo] password for raoul: 
SUBSYSTEMS==usb, ATTRS{idVendor}==16d0, ATTRS{idProduct}==09a0, SYMLINK+=mooltipass TAG+=uaccess, TAG+=udev-acl
moolticute: symbol lookup error: /usr/lib/libharfbuzz.so.0: undefined symbol: FT_Get_Var_Blend_Coordinates

```

azubieta:
That lib was marked as "dangerous" in the template script that I was using so I let is as it.
I'll take a more deep look at it.

azubieta:
I found that only the oldest version are dangerous the new ones are actively maintained and supported by Google and Mozilla.
So I remove the exclusion.

azubieta:
Is this fixed?

raoulh:
[...]

PR #198, opened 2018-01-21
Use udev rules from github.com/bobsaintcool/mooltipass-udev/

azubieta:
issue #149

### DIFF
--- a/data/moolticute.sh
+++ b/data/moolticute.sh
@@ -5,7 +5,19 @@
 
 me="Moolticute.AppImage"
 UDEV_RULES_FILE_PATH="/etc/udev/rules.d/50-mooltipass.rules"
-UDEV_RULE='SUBSYSTEMS=="usb", ATTRS{idVendor}=="16d0", ATTRS{idProduct}=="09a0", SYMLINK+="mooltipass" TAG+="uaccess", TAG+="udev-acl"'
+
+UDEV_RULE="
+# udev rules for allowing console user(s) and libusb access to Mooltipass Mini devices
+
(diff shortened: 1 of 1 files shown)


limpkin:
@bobsaintcool : does it look good?

bobsaintcool:
Hi,

I'm not really into AppImage, so I have only a small understanding on
how it works leading eventually to false assumption for my following
statements. Also the following will only address the device detection
problem not the hardfbuzz not the daemon auto start bug.

I've looked a bit on how the AppImage binary is built under the CI and
how it could be run as a user on Linux (here Arch only).

The previous udev rule provided under [data/moolticute.sh](https://github.com/mooltipass/moolticute/blob/master/data/moolticute.sh#L8) was in fact
not working for Arch Linux user, using mine will fix it as long as
there is strange behavior with AppImage and systemd I'm not aware of.

I've also check a little more the [data/moolticute.sh](https://github.com/mooltipass/moolticute/blob/master/data/moolticute.sh#L61) script that is
run at the starting of the AppImage binary. You check that the udev
rules is present with only the fact that your hardcoded path point to
non-empty file which is eventually uber easy to implement but wrong in
at least two cases.
  * You are trying to master where the udev rule should be installed,
  if a user have made it manually or using a packaged version the
  location and filename will probably not be the one you expect, but
  the udev will probably be installed and correct.
  * If there is a file at the hardcoded location with complete
  garbage, the check the AppImage udev rule installation will be
  bypassed and the app wont work.

I guess using my rule will solve the #149 issue but once again in a
very unsatisfied way to me (ie: you are copying by hand the full
content of a git repository). If someday I decide to change and I
might  in the following weeks, you wont have the update.

PR #215, opened 2018-02-02
MPCmd::printCmd, GET_30_FREE_SLOTS, GET_STARTING_PARENT

sbobykin:
Drafts for `GET_30_FREE_SLOTS` and `GET_STARTING_PARENT`
Using of `MPCmd::printCmd` for better debug info.

For the current moment **MMM** is activated in **Moolticute** UI but it's not operational.
The next step (I think) is to support `WRITE_FLASH_NODE`.

### DIFF
--- a/src/MPDevice_emul.cpp
+++ b/src/MPDevice_emul.cpp
@@ -255,6 +255,19 @@ void MPDevice_emul::platformWrite(const QByteArray &data)
         sendReadSignal(d);
         break;
     }
+    case MPCmd::GET_STARTING_PARENT: //0xC9
+    {
+        // draft
+        QByteArray d;
+        d[0] = 2;
+        d[1] = MPCmd::GET_STARTING_PARENT;
+        d[2] = 0x00;
+        d[3] = 0x00;
+
+        d.resize(64);
+        sendReadSignal(d);
+        break;
+    }
       /* MPCmd::READ_FLASH_NODE
                MPCmd::GET_DN_START_PARENT
                MPCmd::GET_STARTING_PARENT
--- a/src/MPDevice_emul.cpp
+++ b/src/MPDevice_emul.cpp
@@ -222,7 +222,7 @@ void MPDevice_emul::platformWrite(const QByteArray &data)
         sendReadSignal(d);
         break;
     }
-    case MPCmd::GET_CUR_CARD_CPZ: /* 0xC2 */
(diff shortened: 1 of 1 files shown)


limpkin:
Hello there!

I'm afraid I can't merge that, as these particular commands are in direct relation with our internal database model.
So either you choose to implement emulation on a credential level (against @raoulh 's advice) and this code is not needed, or you choose to do emulation on a low level and this code is incorrect.

sbobykin:
@limpkin ok. Part of me agree with you)

This pull request reflect development process only. I've sent it with one reason not to do a large pull request in the future. I think that would be great to have separate dev. branch for emulator.

But after your comment I've got a thought. Why not to create separate daemon for emulator needs only?  In this case we need to realize higher level (websockets based) protocol. Daemon can be created using nodejs. It will be faster to develop. And there is no intersections with existing stuff where that's possible to break something. What do you think?

limpkin:
Right, we do accept development process PRs, but only if their content is correct (which is the complete goal of PRs in the first place :) ).
This way we can give you feedback during your development process.

We can create a separate branch, but the rules will still apply. I've therefore no problem if you stay on the master branch.

A separate daemon for emulation mode would be nice... but wouldn't allow other developers to contribute on the daemon itself... only on the GUI :/

sbobykin:
If we need support for `moolticuted` development then we have to go forward with low level emulation like @raoulh suggests. Other solutions can break existing stuff.

The next PR will be a little wider))

sbobykin:
Another idea has come to me. We can create separate daemon then we can resend requests from `moolticuted` into this daemon.

limpkin:
selecting between which requests are handled by moolticuted and that other daemon may be tricky though no?

sbobykin:
I think that's not so tricky. But there is another problem) If we resend requests then we loose ability to handle `MPDevice`.

So low level emu. wins. @raoulh is right!

If we have low level emu. then we can change almost all the parts in `moolticuted`

sbobykin:
@limpkin BTW

For me that's vary impotent to be in agreement with you in the way how to develop emu. cause I need your help, suggestions and advises.

If you unsure about low level realization then I'd like to continue this discussion because that's a chance to invent better solution.

And may be that's better to move this talk into IRC?

What do you think?

limpkin:
IRC is better yes :)

PR #218, opened 2018-02-07
Qt 5.10: private/qwinoverlappedionotifier_p.h: No such file or directory #201Issue 201

Sriep:
Qt removed qwinoverlappedionotifier_p.h from the Qt5.10 SDK. This fix adds the files qwinoverlappedionotifier_p.h, renamed qwinoverlappedionotifier.h and qwinoverlappedionotifier.cpp to the moolticute code base. Files downloaded from Qtserialport GitHub archive https://github.com/qt/qtserialport/tree/5.10/src/serialport.

These files are only included in the project from Qt 5.10 onwards.

### DIFF
--- a/daemon.pro
+++ b/daemon.pro
@@ -23,10 +23,11 @@ win32 {
 }
 
 win32 {
+    #Private header qwinoverlappedionotifier_p.h removed from Qt5.10 SDK
+    #so need to include files explicitly.
     greaterThan(QT_MAJOR_VERSION, 4):greaterThan(QT_MINOR_VERSION, 9) {
         SOURCES += src/qwinoverlappedionotifier.cpp
         HEADERS += src/qwinoverlappedionotifier.h
-        message (included qwinoverlappednotifier)
     }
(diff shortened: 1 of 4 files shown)


daemon.pro
```diff
@@ -1,7 +1,7 @@
 QT       += core network websockets widgets
 QT       -= gui
 
-#Wee need that for qwinoverlappedionotifier class which is private
+#We need that for qwinoverlappedionotifier class which is private
 win32: QT += core-private
```
raoulh: Please put this into an else {} block after the greaterThan(...)

We don't need core-private anymore in qt5.10

Sriep: qwinoverlappedionotifier.h uses <QtCore/private/qglobal_p.h>. 
qwinoverlappedionotifier.cpp uses <private/qobject_p.h> and <private/qiodevice_p.h>

So we still need core-private, or explicitly include these three files and any private files they need,

raoulh: Good point, it's ok for me then.

PR #220, opened 2018-02-09
Add a button to clear the credential filter in the MMM view

azubieta:


limpkin:
@azubieta 

I'm not getting any icon in the search field.
However, when I'm hovering the place where it should be, i get that:

image

PR #224, opened 2018-02-16
Fix erase credentials filter button

azubieta:


limpkin:
@azubieta : when hovering the button the mouse cursor is still in "text editing mode" and doesn't now show the pointing icon... do you have the same behavior?

PR #234, opened 2018-03-02
Disconnect from progressChanged until the credentials changes are completed.

azubieta:
Please test it,
I'm not sure if it's already implemented the defaultProgressCb when changing the credentials passwords.


limpkin:
ping @raoulh for checks

PR #259, opened 2018-04-17
Added a udev rule and a .desktop file to the Debian package

tomasz-janiczek:


limpkin:
@tomasz-janiczek : instead of hardcoding the udev rule into the repo, could you modify the build script to fetch the udev rule from the udev rule repo?

PR #261, opened 2018-04-22
add SysV init script

wirr00:
Add init script for SysV init based systems

### DIFF
--- a/sysvinit/moolticuted
+++ b/sysvinit/moolticuted
@@ -10,32 +10,32 @@
 
 NAME=moolticuted
 DESC="Mooltipass device management daemon"
-DAEMON=/usr/bin/${NAME}
-DEFAULTS=/etc/default/${NAME}
-PIDFILE=/var/run/${NAME}.pid
+DAEMON="/usr/bin/${NAME}"
+DEFAULTS="/etc/default/${NAME}"
+PIDFILE="/var/run/${NAME}.pid"
 
-[ -x "$DAEMON" ] || exit 0
+[ -x "${DAEMON}" ] || exit 0
 
 OPTIONS=""
-[ -r "$DEFAULTS" ] && . "$DEFAULTS"
+[ -r "${DEFAULTS}" ] && . "${DEFAULTS}"
 
 . /lib/lsb/init-functions
 
 f_start () {
-    log_daemon_msg "Starting $DESC" "$DAEMON"
-    start-stop-daemon --start --quiet --oknodo --background --exec "$DAEMON" --make-pidfile --pidfile "$PIDFILE" -- $OPTIONS
+    log_daemon_msg "Starting ${DESC}" "${DAEMON}"
+    start-stop-daemon --start --quiet --oknodo --background --exec "${DAEMON}" --make-pidfile --pidfile "${PIDFILE}" -- ${OPTIONS}
     log_end_msg $?
     }
 	
 f_stop () {
-    log_daemon_msg "Stopping $DESC" "$DAEMON"
-    start-stop-daemon --stop --quiet --oknodo --retry 2 --exec "$DAEMON" --remove-pidfile --pidfile "$PIDFILE"
+    log_daemon_msg "Stopping ${DESC}" "${DAEMON}"
+    start-stop-daemon --stop --quiet --oknodo --retry 2 --exec "${DAEMON}" --remove-pidfile --pidfile "${PIDFILE}"
     log_end_msg $?
     }
 	
 f_reload () {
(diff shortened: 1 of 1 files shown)


limpkin:
pinging @raoulh  and @bobsaintcool

bobsaintcool:
Hi,

Thanks for the contribution, although not sure sysvinit script has to
be maintained?

Follow are some notes I may have on your PR.

  ##### X-Start-Before
```sh
+# X-Start-Before:    kdm gdm3 xdm lightdm
```
Does this is really needed? Seems hard to me to list here all the
things that could starts X (or wayland?). Also, as moolticute is build
with bunch of stuff from QT if pretty sure it not gonna works without
an graphical session running.

  ##### Run-Level
```sh
# Default-Start:     2 3 4 5
```
I'm not sure considering RL 2 (multi-user without network) will works
either with moolticute as it spawn a lot of sockets. Have you try
this? If not work, may we consider moving RL in the Default-Stop.

  #####  Functions
All along the code you are not using any functions, functions are
good, please consider using it.

  #####  Pidfile/lockfile
No pidfile nor lockfile are used within this script, is this needed?
In the case multiple moolticuted are run on a single node at the same
time, should the daemon crash by itself or the init to do the job?

  #####  sleepy
```sh
+    sleep 1
```
Please no sleep. I don't see any reason one will sleep and in
particular here.

  #####  Stop on error
```sh
+set -e
...
+set +e
```
What are you trying to do here? What the goals?

  #####  Comments
There are some comments that are not really useful as its explain
what simple code means.
```sh
# Check for daemon presence
```
```sh
# Include moolticuted defaults if available
```
```sh
+# Get lsb functions
```

  #####  Daemon name as a variable
I will pick to have a variable name holding the name of the daemon and
use this variable in other part of the program.

```sh
NAME=moolticuded
DAEMON="/usr/bin/${NAME}"
DEFAULTS="/etc/default/${NAME}"
```

And so...

  #####  Quoting
Some variable are quoted but they should not:
```sh
+DAEMON="/usr/bin/moolticuted"
+DEFAULTS="/etc/default/moolticuted"
```

But some are not but they should be
```sh
+  $0 stop
+  sleep 1
+  $0 start
```

  #####  Description
```sh
+# Short-Description: Start the mooltipass password mgmt daemon
```
I disagree on the statement password mgmt. First because not sure
abbreviation must be used here and also because it's not only a
password management daemon but a device management.

  #####  Restart
```sh
+  restart)
+    $0 stop
+    sleep 1
+    $0 start
```

Please don't fork for restart, just write and call stop and start
functions.

  #####  Helpers
```sh
+    log_success_msg "Usage: /etc/init.d/$DAEMON {start|stop|restart|reload|force-reload|status}"
```
This won't work as you think it will. See daemon variable above.

wirr00:
Thanks for the competent and quick reply. As you may have guessed I'm not really an expert on this. This is the init script as it works for me. I will happily rewrite the script following your input. I will post an updated version in a couple of days.

raoulh:
I agree with @bobsaintcool , do we need to support this? All actual distro have switched to systemd.

@wirr00 What distro are you using exactly?

wirr00:
@bobsaintcool I've tried to rewrite the scripts per your recommandations.

**X-Start-Before**
I've quickly tested SSH login using mc-agent and moolticuted on a terminal without graphical session: it seemed to work just fine. Nevertheless I agree it's impossible to list each and every DM. Line removed.

**Run-Level**
A simple test revealed moolticuted doesn't work in RL2; systemd unit file specifies multi-user as target therefore I assume starting moolticuted at RL3 would correspond.

**Functions**
Good point, done.

**Pidfile/lockfile**
I don't know how to handle multiple, manually started instances of the daemon in an init script. Introduced a pidfile nevertheless.

**Sleepy; Stop on error; Comments**
Originated from an LSB init template; Removed.

I've tried to address the rest of your suggestions in my latest commit. Let me know if I should make any other improvements.

bobsaintcool:
Thanks for the feedbacks.

There are still some new little things:

### Quotting
Eventually, introducing variable is a great addition to your script, however I have spotted two situations:

You are now using dynamic path (wtr to variable expansion), thus it will be nice in this case to have quotes arround your path:
```sh
DAEMON="/usr/bin/${NAME}"
```

Also, you have two different style in a single file, some variable are preprended with ${}, where some other only with $. In the provided script both will work but I think having the same scheme all along the file will be better. Pick your favorite.

bobsaintcool:
@raoulh sounds good to me.

raoulh:
thanks

PR #265, opened 2018-04-25
Finished Hungarian Translation

TheRealSoret:
Some preposition problems still exist in translation that I'll try fix once my device arrives.


raoulh:
thanks

PR #276, opened 2018-05-03
Put items on Settings tab into a grid layout

### LINKED ISSUE #69: Minor gui issues
In main window of gui application "Synchronisation" should be "Synchroni**Z**ation", in case the language is English of course.

On settings page in fullscreen mode at display with resolution 1920x1080 the distance between labels and controls is too big (in my opinion):
settings

asashnov:
Closes #69


raoulh:
It does not look good like that, sorry.

PR #278, opened 2018-05-12
209 reset card button when unknown card is detected

asashnov:


limpkin:
getting a warning during compilation: 

> C:\Users\XXXX\temp\moolticute-209-reset-card-button-when-unknown-card-is-detected\src\MPDevice.cpp:6892: warning: unused parameter 'cbProgress' [-Wunused-parameter]
 void MPDevice::resetSmartCard(MPDeviceProgressCb cbProgress,

change the box title from "Uknown card inserted" to "Reset Card"

remove the text and put it in a **tooltip**. Also change the text that way:
"When unknown card is inserted" to "When an unknown card message is displayed"
"in currently connected Mooltipass" to "in your Mooltipass device"
"or use this card..." to "or may use this card in..."
"before reset a card" to "before resetting a card"

Instead of this lengthy test, put the following one: "Erase Inserted Card"

                                                  ^

raoulh:
thanks

PR #287, opened 2018-06-08
fix appearing of monitored backup save requests

asashnov:
Introduce DbMasterController  which is a coordinator for DbExportsRegistryController and DbBackupsTrackerController as they both tried to use the same MainWindow::promptWidget.
This class prevents situations when one class shows prompt for backup a database
and another immediately hides it.


limpkin:
excellent job!

PR #288, opened 2018-06-09
Clang format

netromdk:
After reading the [coding rules](https://mooltipass.github.io/minible/coding_rules), I tried specifying it by way of a .clang-format file to make it easier to conform with the rules.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,68 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: true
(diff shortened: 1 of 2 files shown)


limpkin:
aha, these are the coding rules for the new mini firmware, even though it'd be nice if they could be applied to this repo.  
@raoulh what do you think?

raoulh:
I already tried clang-format, but it misses some options that are automatically done by qtcreator. So i'ts not a good idea, at least for now.

Also those coding rules are for the firmware, C++ ones for moolticute needs some more constrained rules.

PR #289, opened 2018-06-09
Fix warnings

netromdk:
Fixed some warnings:
- Unused `this` captured in lambda
- Added missing `override` keywords
- Marked `this` as unused on macOS since it is used on Windows and Linux

### DIFF
--- a/src/AppGui.cpp
+++ b/src/AppGui.cpp
@@ -139,6 +139,7 @@ bool AppGui::initialize()
                 mainWindowHide();
         }
 #else
+        Q_UNUSED(this)
         Q_UNUSED(reason)
 #endif
     });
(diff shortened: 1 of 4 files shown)


limpkin:
LGTM.
@raoulh can you review and merge?

raoulh:
Thanks

netromdk:
Thanks guys.

limpkin:
thank you @netromdk !

BTW do you want an IRC server access with its companion android app?

netromdk:
You're welcome, @limpkin 

I actually am an Apple man, so to speak. But do you have another IRC channel than FreeNode or a bouncer service maybe? I was actually thinking whether you guys would like to have a Slack team because it's easily available on a lot of systems? :)

limpkin:
@netromdk actually I was offering you a bouncer aha.
Does iOS have a quassel client?

netromdk:
Ah, thanks. Hm it seems there is an iQuassel app. I'd like the bouncer. Thanks!

PR #292, opened 2018-06-11
Capture what is needed instead of everything in lambdas

netromdk:
In relation to talk about memory leaks (on IRC), I looked into what was captured in lambdas in "src/MPDevice.cpp" particularly. The leaks pertain to the fact that lots of the lambdas share multiple copies of data and state, which is compounded into multiple deeper levels. The depth is often a function of the size of the credentials database.

Reasons for not using `[=]`:
- It is unclear exactly what is captured
- In most cases more than necessary is captured, and thus also copied

Notes on the expanded capture list:
- It is longer to write but everyone (also new contributors) knows what data is needed for the scope
- Only what is needed is accessible, and in some cases only as references to avoid copies

I wanted to continue to expand captures in other files but decided that it was better that you guys agreed first. :)

Additional thoughts: In relation to fixing the memory leaks, the copying should try to be avoided by keeping state in MPDevice or by having additional job classes that each maintain only what is necessary until finished - without need for duplicates. The expansions of the lambda captures tries to illuminate what data is necessary for future changes like those.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -3318,7 +3318,7 @@ bool MPDevice::generateSavePackets(AsyncJobs *jobs, bool tackleCreds, bool tackl
     progressCurrent = 0;
     progressTotal = 0;
 
-    auto dataWriteProgressCb = [&](void)
+    auto dataWriteProgressCb = [this, cbProgress](void)
     {
         QVariantMap data = {
             {"total", progressTotal},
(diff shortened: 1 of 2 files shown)


limpkin:
very nice! did you do some kind of testing to make sure nothing was broken?

netromdk:
For each case, I did a compilation test. Afterwards, I ran the daemon and
app with the Mooltipass device connected, and tried the functionality. Due
to the compilation succeeding, we know all data is available and
unfortunately nothing could be made a reference so that means that what was
copied before is copied now. Just only what needed copying.

But there were some cases where nothing needed to be captured. 😊

netromdk:
I also ran the tests but they don't cover everything unfortunately.

raoulh:
thanks

limpkin:
Nice job!

PR #294, opened 2018-06-13
Static analysis

netromdk:
I did some static analysis using Clang's scan-build and CppCheck.

### DIFF
--- a/src/AnsiEscapeCodeHandler.h
+++ b/src/AnsiEscapeCodeHandler.h
@@ -30,10 +30,20 @@ namespace Utils {
 class FormattedText {
 public:
     FormattedText() { }
-    FormattedText(const FormattedText &other) : text(other.text), format(other.format) { }
+
+    FormattedText(const FormattedText &other) {
+        *this = other;
+    }
+
     FormattedText(const QString &txt, const QTextCharFormat &fmt = QTextCharFormat()) :
         text(txt), format(fmt)
     { }
+
+    FormattedText &operator=(const FormattedText &other) {
+        text = other.text;
+        format = other.format;
+        return *this;
+    }
 
     QString text;
     QTextCharFormat format;
--- a/src/AnsiEscapeCodeHandler.h
+++ b/src/AnsiEscapeCodeHandler.h
@@ -31,7 +31,8 @@ class FormattedText {
 public:
     FormattedText() { }
 
-    FormattedText(const FormattedText &other) {
+    FormattedText(const FormattedText &other)
+    {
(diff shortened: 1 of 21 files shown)


netromdk:
For some reason the OSX Travis build fails to download its Qt dependency.. And then takes more than 10 minutes without output so the build fails. I will try to restart it again again.

netromdk:
Came through finally. :)

src/AnsiEscapeCodeHandler.h
```diff
@@ -30,11 +30,21 @@ namespace Utils {
 class FormattedText {
 public:
     FormattedText() { }
-    FormattedText(const FormattedText &other) : text(other.text), format(other.format) { }
+
+    FormattedText(const FormattedText &other) {
```
raoulh: } on newline please, to follow the project code style

src/AnsiEscapeCodeHandler.h
```diff
@@ -30,11 +30,21 @@ namespace Utils {
...
+        *this = other;
+    }
+
     FormattedText(const QString &txt, const QTextCharFormat &fmt = QTextCharFormat()) :
         text(txt), format(fmt)
     { }
 
+    FormattedText &operator=(const FormattedText &other) {
```
raoulh: same

src/AppDaemon.cpp
```diff
@@ -36,6 +38,7 @@ AppDaemon::AppDaemon(int &argc, char **argv):
 
 bool AppDaemon::initialize()
 {
+    assert(!localLogServer);
```
raoulh: use Q_ASSERT instead of assert

src/zxcvbn-c/zxcvbn.c
```diff
@@ -638,9 +638,9 @@ static void DoDictMatch(const uint8_t *Passwd, int Start, int MaxLen, DictWork_t
     for(Len = 0; *Passwd && (Len < MaxLen); ++Len, ++Passwd)
     {
         uint8_t c;
-        int w, x, y, z;
```
raoulh: Actually this is not our code, I wouldn't touch it

src/AnsiEscapeCodeHandler.h
netromdk: Sorry about that.

src/AppDaemon.cpp
netromdk: Will do.

raoulh:
Thanks you for your work!

PR #296, opened 2018-06-15
Lock device when computer is locked, sleeping, shutting down, or logging off

netromdk:
I've implemented a `SystemEventHandler` that currently listens for screen lock system events on Mac and Windows. It then emits `screenLocked()` on that occasion.

The main window UI was extended with a check box for enabling this behavior. When `screenLocked()` is caught it will call `WSClient::sendLockDevice()` that sends JSON and in `WSServerCon` calls
```cpp
mpdevice->sendData(MPCmd::Command::LOCK_DEVICE, cb);
```

Toggling the check box will set/change a bool via `QSettings` for the key `"settings/LockDeviceOnSystemEvents"`, which is used for the state when MC starts.

As @raoulh mentioned, I waited with Linux support until this feature has matured.

I also added events for system sleep, shutdown, and user log off.


src/MacSystemEvents.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SRC_MACSYSTEMEVENTS_H
+#define SRC_MACSYSTEMEVENTS_H
+
+#ifdef __cplusplus
```
raoulh: Are those guard really needed?

src/MainWindow.cpp
```diff
@@ -492,6 +492,14 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
     connect(ui->checkBoxBoot, SIGNAL(toggled(bool)), this, SLOT(checkSettingsChanged()));
     connect(ui->checkBoxTuto, SIGNAL(toggled(bool)), this, SLOT(checkSettingsChanged()));
 
+#ifdef Q_OS_LINUX
+    ui->checkBoxLockDevice->setChecked(false);
+    ui->checkBoxLockDevice->setEnabled(false);
```
raoulh: I would prefer to hide the checkbox on linux for now

src/MainWindow.cpp
```diff
@@ -508,6 +516,17 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
     //Setup the confirm view
     ui->widgetSpin->setPixmap(AppGui::qtAwesome()->icon(fa::circleonotch).pixmap(QSize(80, 80)));
 
+#if defined(Q_OS_MAC) || defined(Q_OS_WIN)
+    connect(&eventHandler, &SystemEventHandler::screenLocked, this, [this]
+    {
+        const bool exec = ui->checkBoxLockDevice->isChecked();
+        if (exec && wsClient->get_status() == Common::Unlocked) {
```
raoulh: { on newline please

src/SystemEvent.h
```diff
@@ -0,0 +1,9 @@
+#ifndef SRC_SYSTEMEVENT_H
+#define SRC_SYSTEMEVENT_H
+
+enum SystemEvent
```
raoulh: Could you move this single value enum in SystemEventHandler.h, having a .h only for 1 enum is not so nice

src/WSServerCon.cpp
```diff
@@ -596,6 +596,11 @@ void WSServerCon::processMessage(const QString &message)
         }
         );
     }
+    else if (root["msg"] == "lock_device")
+    {
+        const auto cb = [](bool, const QByteArray &, bool &){};
+        mpdevice->sendData(MPCmd::Command::LOCK_DEVICE, cb);
```
raoulh: This is not what you want to do. You need to add a new function in MPDevice that will create a new AsyncJob and put it onto the job queue. Directly sending a data packet like that will bypass the current running queue of command and can mess with the communication at some point

src/SystemEvent.h
netromdk: I have another branch which adds three values to this enum: `LOGGING_OFF`, `GOING_TO_SLEEP`, and `SHUTTING_DOWN`.

src/MacSystemEvents.h
netromdk: You're right, I actually don't. :)

src/MainWindow.cpp
netromdk: I can do that.

src/WSServerCon.cpp
netromdk: That makes sense. Will change it into that approach.

netromdk:
@raoulh Thanks for the review. I've made changes in relation to your comments.

netromdk:
@raoulh btw I don't like the timer approach either but the Windows API deprecated and removed the [WM_WTSSESSION_CHANGE](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383828(v=vs.85).aspx) (with `wParam==WTS_SESSION_LOCK`) after Windows 2000. I tried it out anyway and it indeed doesn't get fired.

src/WSServerCon.cpp
```diff
@@ -596,6 +596,19 @@ void WSServerCon::processMessage(const QString &message)
         }
         );
     }
+    else if (root["msg"] == "lock_device")
+    {
+        mpdevice->lockDevice([this, root](bool success, QString errstr)
+        {
+            if (!success) return;
```
raoulh: if !success you need to send back sendFailedJson(root, errstr); before returning.

src/WSServerCon.cpp
```diff
@@ -596,6 +596,19 @@ void WSServerCon::processMessage(const QString &message)
...
+            QJsonObject ores;
+            QJsonObject oroot = root;
+            ores["success"] = "false";
```
raoulh: here it should be = "true"

src/WSServerCon.cpp
```diff
@@ -596,6 +596,19 @@ void WSServerCon::processMessage(const QString &message)
...
+            ores["errstr"] = errstr;
```
raoulh: no errstr because we are in the success case

src/WSServerCon.cpp
netromdk: The point was the if it works then it's evident by the fact that the device's status changes to locked. But sure, can send back response in every case.

src/WSServerCon.cpp
netromdk: Not when it is not a success.

src/WSServerCon.cpp
netromdk: Haha.. I actually meant `if (success) return`..

PR #300, opened 2018-06-19
Save and restore main window geometry and state

netromdk:


netromdk:
@raoulh maybe you can restart the linux builder? It failed to install dependencies.

limpkin:
@netromdk did we lose you on irc?

limpkin:
@netromdk that's odd... can you try firefox?

PR #301, opened 2018-06-20
Proper ⌘+Q behavior on Mac

netromdk:
Two behaviors can be configured in app settings:
1. Close App -> Correctly closes down app
2. Hide Window -> Hides window and doesn't close the app

Since no File->Quit menu action was defined before, ⌘+Q would abruptly terminate the app without
calling all deleters as expected.


limpkin:
@netromdk IRC still working for you? :)

PR #321, opened 2018-07-27
318 odd export behavior

### LINKED ISSUE #318: Odd Export Behavior
Steps to reproduce:
- export db to test.bin
- export db to test2.bin, with generate mooltiapp compatible file ticked
- check that both contents are different
- set monitored file to test.bin
- export db to test2.bin, with generate mooltiapp compatible file ticked
- **notice that the contents of test.bin have been changed in the same way as test2.bin**
- close moolticute
- **get a crash**
- launch moolticute
- set monitored file as test.bin
- add a credential
- click "yes" when asked to update monitored file
- **notice that test.bin isn't using the same encryption as before**

asashnov:
fix #318: fix WSClient::dbExported signal handling

Use the signal consistently: only a part of the program which requested an export should give the export results.

Fix typo SympleCrypt => SimpleCrypt to use 'SimpleCrypt' encryption instead of 'none' in periodic and auto-prompted exports.

### DIFF
--- a/src/DbBackupsTrackerController.cpp
+++ b/src/DbBackupsTrackerController.cpp
@@ -170,8 +170,11 @@ void DbBackupsTrackerController::exportDbBackup()
         format = "SympleCrypt";
     }
 
-
     window->wantExportDatabase();
+
+    // one-time connection, must be disconected immediately in the slot
+    connect(wsClient, &WSClient::dbExported, this, &DbBackupsTrackerController::handleExportDbResult);
+
     wsClient->exportDbFile(format);
 }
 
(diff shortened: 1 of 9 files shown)


src/DbBackupsTrackerController.cpp
```diff
@@ -160,18 +160,13 @@ void DbBackupsTrackerController::askForExportBackup()
 
 void DbBackupsTrackerController::exportDbBackup()
 {
-    QString format;
```
limpkin: are you discarding the file format set by the user when the user is clicking the export button? 
Because this is is not what I asked.... this code is correct, you may simply correct the typo

asashnov: not exactly. class DbBackupsTrackerController initiates a backup prompt when DB in a device is more recent than the file.
User choice is safe: it in https://github.com/mooltipass/moolticute/blob/master/src/MainWindow.cpp#L1164

limpkin:
awesome job!

PR #333, opened 2018-09-06
fix AppImage: specify path to our binary

asashnov:
since we don't use AppRun.c anymore (it's deprecated)


raoulh:
This needs to be merged only when the travis scripts are updated to use the new dockers.

Issue #331, opened 2018-08-31
MacOS, "The Moolticute Daemon is Not Running"


simtomas:
Hi,

with the moolticute versions Moolticute-v0.18.7.dmg and Moolticute-v0.20.2.dmg
there is message "A connection to the Moolticute daemon could not be established."
stays for about 1 minute, until standard page with "Devices settings" etc. shows up.

With a old version (Moolticute-v0.12.7.dmg) the startup is instantaneous (i.e. downgrading to 0.12.7 solves the issue).

macOS version: 10.13.6

Here are the logs output when Moolticute is started from terminal:

$ /Applications/Moolticute.app/Contents/MacOS/moolticute
DEBUG: :0 - System locale:  QLocale(Lithuanian, Latin, Lithuania)
INFO: :0 - Trying to set language:  ":/lang/mc_lt.qm"
DEBUG: :0 - Selection changed
DEBUG: :0 - Db tracking is not available for fw < 1.2
WARNING: :0 - Failed to hide dock icon: Error Domain=NSOSStatusErrorDomain Code=-50 "paramErr: error in user parameter list"
INFO: :0 - Running  "/Applications/Moolticute.app/Contents/MacOS/moolticuted"   ()
INFO: :0 - Daemon started
DEBUG: :0 - Card meta data changed  ""  -  0  -  0
INFO: :0 - Checking software updates from  "https://api.github.com/repos/mooltipass/moolticute/releases"
DEBUG: :0 - "https://github.com/mooltipass/moolticute/releases/download/v0.20.2/Moolticute-v0.20.2.dmg"
DEBUG: :0 - Websocket connected
DEBUG: :0 - Sending cache files request
DEBUG: :0 - New message:  {"msg":"mp_connected"}
DEBUG: :0 - New message:  {"msg":"mp_connected"}
DEBUG: :0 - New message:  {"data":"Locked","msg":"status_changed"}
DEBUG: :0 - New message:  {"data":{"parameter":"keyboard_layout","value":146},"msg":"param_changed"}
DEBUG: :0 - New message:  {"data":{"parameter":"lock_timeout_enabled","value":false},"msg":"param_changed"}
DEBUG: :0 - New message:  {"data":{"parameter":"lock_timeout","value":60},"msg":"param_changed"}
DEBUG: :0 - New message:  {"data":{"parameter":"screensaver","value":false},"msg":"param_changed"}
[...]

PR #334, opened 2018-09-07
Fix issue #331 MacOS, "The Moolticute Daemon is Not Running"

deXol:
With QHostAddress::LocalHost there is a 30 secs delay until  we get the QWebSocketServer::newConnection signal, while with Any or LocalHostIPv6 it works instantly.


raoulh:
Thanks

PR #338, opened 2018-09-13
Fix Issue #331 MacOS, "The Moolticute Daemon is Not Running"

deXol:
Communication with browser extension did not worked with LocalHostIPv6, so reverted to using Localhost and Any on Mac. Because on Mac with Localhost option "The Moolticute Daemon is Not Running" message remains for 30 seconds before Device Settings appears.


raoulh:
Thanks

Issue #253, opened 2018-04-03
Feature request: query user to input login if field is empty


limpkin:
When using native integration with browser extensions, if a "store credential" message is sent to moolticute and the login is empty, generate a popup requesting the user to input the login.

limpkin:
implemented

Issue #255, opened 2018-04-09
Feature Request: ask the user to select between subdomain and domain in a popup rather than in the extension


limpkin:
... following the logic to move all notifications into MC

raoulh:
On the MC side, the notifications needs to be handled by using a the OS notification systems. The idea is to integrate well with the underlying OS notifications systems and prevent having weird MC popups all the time. This way, our notifications are non intrusive to the user. Notifications that needs to use that are:

- Browser extensions, they will send a message to MC to display a notification
- Update popups, they should be sent to the system notifications

To implement this we can take a look at those ressources:

- https://github.com/KDE/snorenotify
- https://techbase.kde.org/Projects/Snorenotify/Windows-Toast-Notification

Interestingly on windows they are using a single .exe. 

> "The reason for using a command line application is, that applications enabled for notifications on Windows must be compiled exclusively for Windows 8 or later and written in a way that is not yet supported by Qt."

Snorenotify seems to be too big, so the idea is just to see how they do the backend thing and implement that in moolticute, and support: 

- OSX Notification Center,
- Windows Toast notifs,
- Free Desktop Notifications.

Growl and such are of no use for us anyway.

deXol:
The current behavior, when a subdomain is detected is the following notifications from the browser extension:
![domaincurrent](https://user-images.githubusercontent.com/11043249/45907855-f897bb80-bdf9-11e8-9d26-b03c2ce0d20a.jpg)
I tried Windows Toast Notification and this is possible to display the same message easily with the given binary SnoreToast:
![domainsnoretoast](https://user-images.githubusercontent.com/11043249/45907884-25e46980-bdfa-11e8-9bac-e3b4c0a33b1b.jpg)
However according to the feature request the user should be able to choose between how to store the service: only the domain or the subdomain too. This is not possible to do in SnoreToast, because the API doesn't contains combobox.
I've implemented a demo similar to #339, just to have the logic:
![domainrequest](https://user-images.githubusercontent.com/11043249/45908040-d05c8c80-bdfa-11e8-927a-50180ef85d9f.jpg)
Here we are able to choose between the domain and subdomain.

limpkin:
@deXol : I would go for snoretoast for notifications and dialog for subdomain choice.  
We however now need to make sure the user doesn't get twice the same notification.  
[...]

PR #339, opened 2018-09-15
Implemented issue #253 Feature request: query user to input login if …

deXol:
…field is empty

If there is a new credential in the browser without login name, then daemon sends a signal to GUI for request the login from the user. When the user added his/her login name in the MessageBox, then the GUI sends a message to the daemon with the given login name.

An example:
loginrequesttwitterexample

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1043,9 +1043,8 @@ void MainWindow::wantExitFilesManagement()
 void MainWindow::displayLoginRequestMessageBox(const QString& service, QString& loginName, bool& abortRequest)
 {
     RequestLoginNameDialog dlg(service);
-    dlg.exec();
+    abortRequest = (dlg.exec() == QDialog::Rejected);
     loginName = dlg.getLoginName();
-    abortRequest = dlg.abortRequest;
     if (!loginName.isEmpty())
     {
         qDebug() << "Login name is set: " << loginName;
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -25,6 +25,7 @@
 #include "PasswordProfilesModel.h"
 #include "PassGenerationProfilesDialog.h"
 #include "PromptWidget.h"
+#include "RequestLoginNameDialog.h"
 
 #include "qtcsv/stringdata.h"
 #include "qtcsv/reader.h"
@@ -1039,14 +1040,13 @@ void MainWindow::wantExitFilesManagement()
     updateTabButtons();
 }
 
-void MainWindow::displayLoginRequestMessageBox(const QString& service, QString& loginName)
+void MainWindow::displayLoginRequestMessageBox(const QString& service, QString& loginName, bool& abortRequest)
 {
-    bool ok;
-    QString loginRequestString = tr("Login name for ") + service + ":";
-    loginName = QInputDialog::getText(this, tr("Login Request"),
-                                             loginRequestString , QLineEdit::Normal,
-                                             "", &ok, Qt::WindowStaysOnTopHint | Qt::MSWindowsFixedSizeDialogHint);
-    if (ok && !loginName.isEmpty())
+    RequestLoginNameDialog dlg(service);
+    dlg.exec();
+    loginName = dlg.getLoginName();
+    abortRequest = dlg.abortRequest;
+    if (!loginName.isEmpty())
     {
         qDebug() << "Login name is set: " << loginName;
     }
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -151,6 +151,7 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
     connect(wsClient, &WSClient::wsDisconnected, this, &MainWindow::updatePage);
     connect(wsClient, &WSClient::connectedChanged, this, &MainWindow::updatePage);
     connect(wsClient, &WSClient::statusChanged, this, &MainWindow::updatePage);
+    connect(wsClient, &WSClient::displayLoginRequest, this, &MainWindow::displayLoginRequestMessageBox);
 
     connect(wsClient, &WSClient::memMgmtModeChanged, this, &MainWindow::enableCredentialsManagement);
     connect(ui->widgetCredentials, &CredentialsManagement::wantEnterMemMode, this, &MainWindow::wantEnterCredentialManagement);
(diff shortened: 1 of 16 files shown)


raoulh:
Hi,

I don't really like this.

- First I do have some website where I have no login at all. Right now with this PR, I can't save any password for this website anymore. The only way is to put a random login. But then it enters the login when using the device alone. I mean when using the device alone with a credential that has no login, it only sends the password and do not ask for the login at all which is nice.

- Second, it will break and do nothing if you don't have any GUI running.

For me a "feature" like that should be implemented in the client which would send `the set_credential``at the beginning. Like here it's the extension that would need to ask the user if it's ok for him to use no login for this credential or query one directly.

deXol:
Hello,

I suggested the extension solution too.
But if we stay with this solution, I can add a flag for set_credential to save without a login in this two case.
So if the GUI is not running or the user manually clicked OK on an empty login, I add this flag and process it with empty login as before.

limpkin:
Makes sense to me!

As a reminder, this feature was requested because we want to migrate all
prompts to moolticute as some browsers do not allow prompts

src/MainWindow.cpp
```diff
@@ -1038,6 +1039,36 @@ void MainWindow::wantExitFilesManagement()
...
+{
+    QJsonParseError err;
+    QJsonDocument jdoc = QJsonDocument::fromJson(message.toUtf8(), &err);
+    if (err.error != QJsonParseError::NoError)
+    {
+        qWarning() << "JSON parse error " << err.errorString();
+        return;
+    }
+    QJsonObject rootobj = jdoc.object();
+    QJsonObject o = rootobj["data"].toObject();
```
raoulh: All the json decoding should stay in WSClient. It's part of the json/websocket procotol, it should not be done in mainwindow

deXol:
@raoulh Ok, I will fix these issues and then I create a new bigger dialog with those informations.

I am working on #255 which also requires very similar changes. Then I guess that also requires a new window or should I make this a generic one? (Texts are different there, and combobox instead of input field)

src/MainWindow.cpp
deXol: moved to WSClient

deXol:
Added empty logins, now working on the new dialog.

raoulh:
@deXol #255 needs to be handled using the system notification from the OS. Not using our own dialogs. I will write some details about that on the issue

src/WSServer.cpp
```diff
@@ -101,6 +102,20 @@ void WSServer::notifyClients(const QJsonObject &obj)
...
 }
 
+void WSServer::notifyGUI(const QString &message, bool &isGuiRunning)
+{
+    for (auto it = wsClients.begin(); it != wsClients.end(); ++it)
+    {
+        //Notify GUI
+        if (it.key()->resourceName().contains("localhost") && !(it.key()->origin().contains("extension")))
+        {
+            it.value()->sendJsonMessage(message);
```
raoulh: `isGuiRunning` should be set to true here, it's bad to let it be initialized to a value from the caller of this function.

limpkin:
Quick note for the future: imagine the following scenario:
- user visits a website for which he has credentials, but is only requested
his password (eg Google)
- the password submit is detected by the extension, which requests the user
for the login

For this pr as well as for the upcoming subdomain one, I therefore suggest
to not request the user anything if a previous credential request was
previously approved for the same complete URL.

src/WSServer.cpp
deXol: thanks, I forgot to add it there.

deXol:
> Quick note for the future: imagine the following scenario: - user visits a website for which he has credentials, but is only requested his password (eg Google) - the password submit is detected by the extension, which requests the user for the login For this pr as well as for the upcoming subdomain one, I therefore suggest to not request the user anything if a previous credential request was previously approved for the same complete URL. On Mon, Sep 17, 2018, 13:10 mathieu.stephan@gmail.com < mathieu.stephan@gmail.com> wrote:
> […](#)

For a few days I have access only to a Mac, I tried google (gmail) with Safari, and if I try to login the daemon is not receiving the set_credential message. If I try to choose the login and password field manually with the extension, then I cannot do that neither, because only the site darken and cannot click anywhere.
I will try this on Windows in a few days.

deXol:
Added the new login request dialog:
screen shot 2018-09-18 at 9 55 36 pm
Now we can add empty login name, it will accepted.
And if we choose cancel it will cancel the storage request.
If you have any other request regarding the dialog please let me know.

src/RequestLoginNameDialog.cpp
```diff
@@ -0,0 +1,37 @@
...
+    QDialog(parent),
+    ui(new Ui::RequestLoginNameDialog)
+{
+    ui->setupUi(this);
+    setFixedSize(size());
+    setWindowTitle(tr("Request Login Name"));
+    ui->loginLabel->setTextFormat(Qt::RichText);
+    ui->loginLabel->setText(tr("Login name for ") + "<b>" + service + "</b>:");
+    const QString informationText = "<h2><b>"
+                                  + tr("A credential without a login has been detected.")
```
raoulh: Change the text to:

`<h2><b>A credential without a login has been detected.</b></h2><br><br>You have the possibility to include a login before storing the credential in the device. If you don't need a login, you can ignore the request.`

Also, could you include an `ignore` button in the buttonbox

deXol:
Added text and ignore button:
screen shot 2018-09-19 at 10 32 35 pm
If we click Ignore it is saving the empty login, and click on Cancel or escape is canceling the storage request.

deXol:
> During testing I found an issue on mac.
> If we open any dialog from MainWindow, close it, hide MC and we are trying to show it again, then it will crash.
> It looks independent from this new dialog, because I could reproduce with Setting/Password Profiles... also every time I tried.
> It is crashing in AppGui.cpp:382
> `utils::mac::orderFrontRegardless(win->winId(), true);`

Fixed in the last commit.
According the callstack this problem caused the crash in the following ticket too: https://github.com/limpkin/mooltipass/issues/558

raoulh:
@deXol Please remove the commit "Fixing crash on Mac from orderFrontRegardless" from this PR, and open an other PR for this. It does not belong here. thanks

deXol:
> @deXol Please remove the commit "Fixing crash on Mac from orderFrontRegardless" from this PR, and open an other PR for this. It does not belong here. thanks

reverted it

limpkin:
@raoulh any missing changes?

deXol:
Added extension_version check in set_credential message from the browser extension according to the following comment: https://github.com/mooltipass/moolticute/issues/255#issuecomment-423836246 in #255

src/WSServerCon.cpp
```diff
@@ -225,7 +225,7 @@ void WSServerCon::processMessage(const QString &message)
     {
         QJsonObject o = root["data"].toObject();
         QString loginName = o["login"].toString();
-        if (loginName.isEmpty() && !o.contains("saveConfirmed"))
+        if (loginName.isEmpty() && o.contains("extension_version") && !o.contains("saveConfirmed"))
```
raoulh: I'm against this. First it does not belong to this PR, second if we use the `mc-cli` or the API to set an empty login we do not get the prompt.

limpkin: first point: agreed, though we eventually need to integrate it somewhere
second point: I had forgotten about mc-cli. Could we add a "mc_cli_version" key and do a OR on that condition?

raoulh:
Lets merge this now. Thanks @deXol

PR #341, opened 2018-09-21
Fixing crash on Mac from orderFrontRegardless

deXol:
orderFrontRegardless was responsible on Mac to show the MainWindow on the top after it was hidden. However it crashed very often and even if it was not crashing, it hasn't brought the window in front.
It crashed because the used winId is changing during runtime, so it is dangerous to use it:
http://doc.qt.io/archives/qt-4.8/qwidget.html#winId
Instead of orderFrontRegardless the raise and activateWIndow call is bringing the MainWindow to the top level after clicking show without a crash.

### DIFF
--- a/src/AppGui.cpp
+++ b/src/AppGui.cpp
@@ -378,10 +378,14 @@ void AppGui::mainWindowShow()
     }
 
     win->show();
+#ifdef Q_OS_MAC
+    win->raise();
+    win->activateWindow();
+#endif
+
     showConfigApp->setText(tr("&Hide Moolticute App"));
 #ifdef Q_OS_MAC
     utils::mac::hideDockIcon(false);
-    utils::mac::orderFrontRegardless(win->winId(), true);
 #endif
 }
 


limpkin:
@raoulh good to merge?

raoulh:
No, not good at all. There was a reason at first why we did not use `raise()` and `activateWindow()`. It does not work correctly in all case. Wee need to fix the `orderFrontRegardless` crash, not just bypass it...

raoulh:
I don't remember actually. I did that 2 years ago and had problems.
So it would be better to fix our function. Actually there is no check for `window `or `widget` beeing NULL...

limpkin:
@raoulh could you let @deXol know on which platform you had problems so he can try to replicate the bug?

raoulh:
@limpkin that part of the code is macOS only... and @deXol knows that ;-)

Anyway we need to fix our `orderFrontRegardless` code to not crash anymore.

deXol:
> 
> 
> I don't remember actually. I did that 2 years ago and had problems.
> So it would be better to fix our function. Actually there is no check for `window`or `widget` beeing NULL...

Just asking that because it could have been a Qt issue with Mac, which have been fixed since then. Because I tested it with several scenario and never crashed with `raise()` and `activateWindow()`.
On the other side `orderFrontRegardless` is not only crashing, but it doesnt even work, if there is no crash. (MC client remains in the background)
There is a NULL check for `win` in line 373.

raoulh:
Ok, maybe you're right. Lets keep the raise() and activateWindow() for now, and see if we have problems with that in the future.

limpkin:
woohoo!

PR #342, opened 2018-09-23
#255 Develop issue255 domain selection

deXol:
Added the dialog for choosing between domain and subdomain if subdomain is detected:
domainrequest

Also added the check for extension_version in set_credential, so it can be tested only if the set_credential message from extension contains "extension_version" data.


deXol:
Merged to https://github.com/mooltipass/moolticute/pull/348

Issue #343, opened 2018-09-24
Notification system next steps


limpkin:
- @deXol to investigate if #339 & #342 can realistically be implemented using native OS notification APIs
- future "standard" notifications (device disconnected, please unlock device...) to be implemented using native OS notification APIs
- mc-cli to add a "mc_cli_version" key to the get credential message

raoulh:
Keep in mind that we need:

- Windows notifications using toast api for windows >= 8
- Something else on windows 7? (that's a question to resolve)
- macOS notifications
- Linux freedesktop notifications

And with the following features:
- Text only
- Question
- Prompt with selectable items (subdomain thing)

deXol:
**Windows**:

1. **Toast notifications api:**
https://blogs.msdn.microsoft.com/tiles_and_toasts/2015/07/08/toast-notification-and-action-center-overview-for-windows-10/
According to this **ToastGeneric** contains the features which we need (editbox, combobox), but **only Windows 10 support**s them: https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/send-local-toast-desktop-cpp-wrl#handling-older-versions-of-windows
I couldn't find a lib for c++ which implemented these functions, the closest was WinToast, but even that only supports adding buttons to the notifications.
So we should implement the COM calls according to this page for the missing features: 
https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/send-local-toast-desktop-cpp-wrl

2. **WinToast:**
https://github.com/mohabouje/WinToast
	- Using WinRT (WRL), the above mentioned Toast notifications api
	- Only works with MSVC2015 compiler
	- Implemented with win32 api COM calls
	- Can add buttons
	- No inputbox or combobox (but possible to extend it with these)
![wintoastexample](https://user-images.githubusercontent.com/11043249/45977251-4ce4aa80-c049-11e8-804f-83eb155e71b5.jpg)

	https://stackoverflow.com/questions/28609521/windows-10-toast-notifications-desktop-application
	
3. **SnoreToast**:
https://github.com/KDE/snoretoast
	- Binary file
	- Calling the binary with command line argument
	- Only for notifications
![domainsnoretoast](https://user-images.githubusercontent.com/11043249/45977517-2ffca700-c04a-11e8-854a-cd4f04d56dcc.jpg)


	
**Mac**:

1. **OS_X_Notification_Center**
[...]

PR #348, opened 2018-10-01
Develop343 adding system notifications

deXol:
#343 
**[for review]**
- Extended SnoreToast on a fork:
   https://github.com/deXol/snoretoast

- Added the SnoreToast.exe binary and copy it during build on Windows.
- Created an interface for system nofitifications, which will be implemented on every platform. 
- Implemented the SystemNotification for Windows. Currently the following interfaces are avaliable:
     - **createNotification(title, text)**: Simple notification with two line of text
     - **createButtonChoiceNotification(title, text, buttons)**:  Notification with button options, user choice is handled.
     - **createTextBoxNotification(text, title)**: Notification with a textbox, user input is handled.
     - **displayLoginRequestNotification(service, [out] loginName)**: Displaying the login request notification for requesting the login name, user input is returned in loginName.
- Integrated the empty login handle. On Win10 it is displaying the notification with a textbox and bellow win10 the RequestLoginNameDialog is displayed.
    
emptyloginsnoretoastintegrated

When we are using RequestLoginNameDialog it is displayed on the top of MainWindow. But with notification we can see the MC GUI, however because we invoked the notification from the main thread the GUI is frozen and the user cannot use it until the notification is dissapearing.
Should I change it to start the notification on other thread or they should not use it while the notification is up?
I commented out a solution for handling user interaction from the new thead, but it would require some more changes too.


limpkin:
I don't see why the user would need to use the main MC GUI while leaving the prompt opened...

deXol:
I merged the subdomain selection branch here, and integrated it to system notification:
domainselectionsnoretoastintegrated

Checked why Travis is failing on Unix and found in the pro file every win32 conditional parts are included with every windows's cpp, header and configurations. So it tries to copy SnoreToast.exe, which fails.

limpkin:
@deXol could you also make a setting in the "Settings" tab of moolticute to enable this prompt and therefore allow the user to select between domains? 
By default, it should store (in your example) raffeisen.hu. Sub-domain selection is kind of a power-user thing. As a matter of fact, only chrome extension has this feature.

deXol:
> 
> 
> @deXol could you also make a setting in the "Settings" tab of moolticute to enable this prompt and therefore allow the user to select between domains?
> By default, it should store (in your example) raffeisen.hu. Sub-domain selection is kind of a power-user thing. As a matter of fact, only chrome extension has this feature.

@limpkin Added in this commit: https://github.com/mooltipass/moolticute/pull/348/commits/97164ad9535b9e6f16d66315ea49498df3678433

enablesubdomainselectionsetting

Also added to store without subdomain by default.

limpkin:
great work so far @deXol !

deXol:
Implemented the system notification for **Mac** too.
Empty login notification:
emtylogin

Domain selection notification:
subdomainselection

deXol:
**Unix**
I implemented the freedesktop notification with DBus according the mentioned snorenotify's solution. (https://github.com/KDE/snorenotify/tree/master/src/plugins/backends/freedesktop_backend)
It works well with simple systemnotification:
unixnotification
However if I add buttons for actions, it becomes an ugly alert dialog, because system notification does not support it:
unixnotificationwithaction
There is no option for textboxes at all.

Currently RequestDomainSelectionDialog and RequestLoginNameDialog is used for Unix. If there is a solution for the buttons or textbox I will change it.

limpkin:
awesome work!  
Does that mean that feature request is effectively done?

deXol:
> 
> 
> awesome work!
> Does that mean that feature request is effectively done?

If there is no other solution for Unix, then I will do some minor changes (moving System Notifications related code to a new folder) and test it on every environment today. After that it is done.

deXol:
### Summary:

- **Windows**
    -  Added functionalities to the SnoreToast fork
    -  Using SnoreToast.exe
    -  Empty login and subdomain selection is using System Notification on Win10, RequestDomainSelectionDialog and RequestLoginNameDialog bellow Win10
    - https://github.com/mooltipass/moolticute/pull/348#issue-219359295

- **Mac**
     - Using Mac system notification
     - Empty login and subdomain selection is using System Notification
     - https://github.com/mooltipass/moolticute/pull/348#issuecomment-427149199

- **Unix**
    - Implemented freedesktop with using dbus for simple notification
    - Empty login and subdomain selection is using RequestDomainSelectionDialog and RequestLoginNameDialog, because there were no button or textbox options for System notification
    - https://github.com/mooltipass/moolticute/pull/348#issuecomment-427686343

### Other issues:
- Added an option to Settings for Enable and Disable subdomain selection
- Set domain.tld for default service
- Travis is failing on Linux, because it is trying to copy SnoreToast.exe which is in a win32 configuration block. For me the build was successful on Unix (Ubuntu 14.04), it is maybe some configuration problem on Travis.

raoulh:
Hi,

I finally got some time to review the PR.

First, good work @deXol.

I have some concerns about the main design decisions that you made though. Everywhere you need to somehow block the main loop, either by creating QEventLoop or by using other methods. I really don't like this approach, for many reasons. But first I would prefer to keep the entire notification system asynchronous without timeouts. As such it means sending notifications to the OS, and just sit and wait. If the user click on one of the notif at one point in time (beeing 5s or 2 days should not matter as the notifications are in the OS's queue anyway) it should be handled. The only point is that moolticute should be kept running and if the user exits MC, the notifications are not valid anymore.

After looking into the code, it should be doable to do all async. For example on windows, mc can start as many SnoreToast.exe as per notifications requests needed. For mac/linux the api are already async. It should only be handled correctly in MC to keep track of notifications.

So @deXol could you tell us what you think about doing it the way I explained. Thanks

deXol:
Hello,

I asked the async part in the first comment of the PR and implemented according to the answer.
For windows there is a commented out solution for the async callback. And the other OS are async by default, just added waiting to user interaction for the same behavior.
If we want it to be async, it only needs to save the set_credential message for the system notification id and when the callback arrives send it back with the result.

limpkin:
@deXol the blame is on me and raoul... we didn't think it through at the time when you asked the question :)

deXol:
@limpkin No problem, I can change it to async. :)

raoulh:
Yes, sorry... I did not have much time to review this until today...

deXol:
I implemented the async system notification handling for Windows and added some changes regarding notification id for the SnoreToast fork.

> If the user click on one of the notif at one point in time (beeing 5s or 2 days should not matter as the notifications are in the OS's queue anyway) it should be handled.

However this is an issue for now, because if the notification is disappearing, then a dismiss event is invoked and the event handler is unavailable from that point, but the notification is still in the notification center.
I'm investigating if this is possible in SnoreToast to keep alive the event handler even if the notification is moved to the notification center.

**A question regarding the update dialog:**
Where the system notification with action buttons are not available (e.g.: bellow win10, unix) should I use the current behavior?

deXol:
I experienced on Windows and Mac, if I set the settings to Do Not Disturb, then it is not displaying system notification at all.
This was a problem mainly on Mac, because there are some cases (e.g.: mirroring display) where Do Not Disturb is the default settings.

limpkin:
Win10: what is causing the dismiss event? For some notification I can still click them on the event history (I think...)
I think we can use the current behavior for <win10 & linux. 
@raoulh : what do you suggest for DnD?

deXol:
> Win10: what is causing the dismiss event?

OS sending a ToastDismissedEvent with ToastDismissalReason_TimedOut reason on notification disappearing. 
For DnD it is sending instantly this.

limpkin:
Interesting! Can we detect DND? Can we remove notifications for which we
received the time out?

deXol:
> 
> Interesting! Can we detect DND?

It is sending the same event like when it disappears. Maybe it is somewhere in registry, I will check it.

> Can we remove notifications for which we received the time out?

Yes, this is possible. However it looks like I found a solution, where we have the event handler after the notification is dismissed by the OS, just need few more changes for that.

limpkin:
neat! what's the solution out of curiosity? another event handler for the same functionality?

deXol:
> 
> 
> neat! what's the solution out of curiosity? another event handler for the same functionality?

Yes, using the other (from http://www.naughter.com/toastpp.html) eventhandler completely, which was implemented for textbox interaction.
The problem was with SnoreToast's eventhandler after receiving the dismissed event it terminated.

deXol:
Fixed the issue in the SnoreToast fork. (https://github.com/deXol/snoretoast/commit/e38daf8445e09cf9864a02dac8297489891c8dc5) And added the new binary here.
Now notifications in notification center are working, whenever the user interacts with them:
notificationsinnotcenter

Worth to mention regarding the eventhandler in SnoreToast, it is working like a stack. So e.g.: if I start 3 new notifications and I want to handle them from the notification center, it does not matter which I choose the first notification's eventhandler will process that.
This is handled by using unique ids in the notifications.

deXol:
> Interesting! Can we detect DND?

I found the registry entry for that on windows. Added a function for checking it.

deXol:
> 
> 
> Nice work @deXol
> 
> About the update notification, I would prefer to have a simple notification that tells user about an update available, without any action. Then when user open moolticute it should display a banner on the windows which would prompt user to start the upgrade. This way it's non intrusive. What do you think?

Thanks.

In MC client do you mean the PromptMessage like for DB export?
I've done a little demo for that:
updatenotificationandprompt
I checked this option and I think it could work, however there are some issue.
There was an implementation for the Update dialog, where I displayed the changelog for the newest update and this would not fit in this banner.
Another problem that the dialog is in the Updater project, which is displaying the Update dialog and currently is independent from the GUI, but for this we should pass the mainwindow to the Updater for showing the banner. I do not know if that is a problem or not.
The Update dialog is not async now (cannot use the client when it is displayed), hence there should be more change to handle it async with the new solution.

If these are not a problem I can implement it, but first I think I will do the async System Notification handling for Mac.

raoulh:
Why not adding a changelog button/link inside the banner that opens the changelog if user wants it?

limpkin:
Good idea

deXol:
Had some minor fixes for Unix and fixed travis issue.
Now the system notifications are done.
So I will implement the simple notification for update and banner in mainwindow solution with a link for changelog.

deXol:
I implemented the new Update notification:
updatenotificationwin
If a new version is detected, then a **system notification** is sent and the **banner** is displayed on MainWindow.
The user can **open the changelog** information about this version with the link, **start the updat**e with clicking yes and **hide the banner** with no.
Updates are checked every 30 minutes.

With this I think the PR is finished.

raoulh:
Great work @deXol ! I will do a full review and some test as soon as I can.

deXol:
> 
> 
> compilation warnings on latest qt on win:
> 
> \src\qwinoverlappedionotifier.cpp👎 In member function 'OVERLAPPED* QWinOverlappedIoNotifierPrivate::waitForAnyNotified(QDeadlineTimer)':
> 
> \src\qwinoverlappedionotifier.cpp:319: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
> if (msecs == -1)
> ^

It's trying to compare msecs which is a DWORD (unsigned long long) with -1, which in fact is a problem.
I can fix it, but havent changed the file in this PR.

limpkin:
I'm guessing @raoulh will ask you to make some changes anyway so you may as well change it here :)

It seems since in this PR the "standard" messages are higher though :/ :  

image

limpkin:
I'm guessing @raoulh will ask you to make some changes anyway so you may as well change it here :)

It seems since in this PR the "standard" messages are higher though :/ :  

image

@raoulh : do not forget to modify mc-cli json messages to include "mc_cli_version" : https://github.com/mooltipass/moolticute/pull/348/files#diff-d2c769b75b85702cadd646dcd0917fddL228

limpkin:
I'm guessing @raoulh will ask you to make some changes anyway so you may as well change it here :)

It seems since in this PR the "standard" messages are higher though :/ :  

image

@raoulh : do not forget to modify mc-cli json messages to include "mc_cli_version" : https://github.com/mooltipass/moolticute/pull/348/files#diff-d2c769b75b85702cadd646dcd0917fddL228

limpkin:
I'm guessing @raoulh will ask you to make some changes anyway so you may as well change it here :)

It seems since in this PR the "standard" messages are higher though :/ :  

image

@raoulh : do not forget to modify mc-cli json messages to include "mc_cli_version" : https://github.com/mooltipass/moolticute/pull/348/files#diff-d2c769b75b85702cadd646dcd0917fddL228

limpkin:
I'm guessing @raoulh will ask you to make some changes anyway so you may as well change it here :)

It seems since in this PR the "standard" messages are higher though :/

@raoulh : do not forget to modify mc-cli json messages to include "mc_cli_version" : https://github.com/mooltipass/moolticute/pull/348/files#diff-d2c769b75b85702cadd646dcd0917fddL228

limpkin:
I'm guessing @raoulh will ask you to make some changes anyway so you may as well change it here :)

It seems since in this PR the "standard" messages are higher though :/ : 

image


@raoulh : do not forget to modify mc-cli json messages to include "mc_cli_version" : https://github.com/mooltipass/moolticute/pull/348/files#diff-d2c769b75b85702cadd646dcd0917fddL228

deXol:
Added install SnoreToast on startup for Windows and simple system notifications for credentials detected.

limpkin:
Nearly everything is OK for me!
A few remaining points:
- on one of my computers, it seems DnD isn't detected (called focus assist in my case):
image
- anyway to rename snoretoast into Mooltipass below?
image

deXol:
> 
> 
> Nearly everything is OK for me!
> A few remaining points:
> 
>     * on one of my computers, it seems DnD isn't detected (called focus assist in my case):
>       image
> 

Is HKEY_CURRENT_USER\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Notifications\\Settings\\NOC_GLOBAL_SETTING_TOASTS_ENABLED registry entry set on that computer during DnD?

>     * anyway to rename snoretoast into Mooltipass below?
>       image

I will check this, maybe if I rename the project it could work.

limpkin:
image
... and in mobilityexperience:
image

deXol:
>  ... and in mobilityexperience:

It should be an entry in the Settings if DnD is enabled:
kep

limpkin:
nope, nothing

deXol:
> 
> 
> nope, nothing

Maybe is it in same root, but in HKEY_LOCAL_MACHINE instead of HKEY_CURRENT_USER?

deXol:
I was able to rename SnoreToast in notifications:
kep

kep

Pushed changes to snoretoast fork, added the new binary.

limpkin:
wonderful!

raoulh:
I merged with some minor changes. see https://github.com/mooltipass/moolticute/commit/46a8cea27668eca6ee87d603aaca4305bd510073

Issue #354, opened 2018-10-21
Black systray icon: wrong setting selected by default


limpkin:
Inside moolticute's settings tab, black seems to be selected by default regardless of the current setting

PR #355, opened 2018-10-22
#354 Setting default systray option during first start

deXol:
During first start "settings/systray_icon" QSetting wasn't persisted, but the combobox reads the default value from this QSetting.


deXol:
#354

Issue #349, opened 2018-10-02
No autostart after MC install on windows


limpkin:
confirmed on 2 different computers, one being freshly formatted.
hack that needs to be implemented: MC needs to check for its first boot and enable autostart

limpkin:
fixed

PR #356, opened 2018-10-23
#349 Adding default true to autostart after install

deXol:
This is a very similar issue to #354.
During first start QSettings does not have a value for "settings/auto_start" in the registry, so by default it returns false.
Now this case is handled by setting it to true.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -30,6 +30,8 @@
 #include "qtcsv/stringdata.h"
 #include "qtcsv/reader.h"
 
+const QString MainWindow::AUTO_START_SETTING = "settings/auto_start";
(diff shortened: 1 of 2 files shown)


src/MainWindow.cpp
```diff
@@ -1112,7 +1114,19 @@ void MainWindow::checkAutoStart()
 {
     QSettings s;
 
-    bool en = s.value("settings/auto_start").toBool();
```
raoulh: s.value() supports default value.
Easier to just do:

bool en = s.value("settings/auto_start", true).toBool();

raoulh:
Actually I did the small fix, no need for a pr. thanks

Issue #351, opened 2018-10-13
MMM: make changes to the bottom buttons


limpkin:
Suggestion: when entering MMM (no changes made yet), show a "Exit Credential Management".  
If a modification has been made to any credential, make the 2 current buttons appear.

deXol:
Added change in #358.

During implementation I experienced a weird error.
**Repro**:
1. Go to Credentials
2. Enter Credentials Management Mode with Mooltipass device
3. Add a new credential with "Add or update a credential" section
4. Click Save all changes
5. Mooltipass device asks "Change Password For" if the user discards it or timeouts, then the daemon crashes, the user cannot use MC, until restarting the daemon and the added credential is saved with some invalid password (originaly it was "test" too):
6. ![kep](https://user-images.githubusercontent.com/11043249/47529879-5c098300-d8a9-11e8-97cd-1c038eecadbf.png)

**Occurrence**: every time
**Test on**: Windows, Ubuntu

limpkin:
nice find!
On your current PR I did manage to replicate the crash using a timeout and not the deny...
You may investigate if you have some spare time, I suspect something at a low level is happening.
Line 6834 of mpdevice.cpp is the one you'll want to look at

PR #358, opened 2018-10-25
#351 Adding exit button to Credentials management in clean state

deXol:
Added the new exit button:
mmmexit
It is presented until a change made regarding credentials (added, removed, modified, added to fav), after that the original save and discard is displayed.


limpkin:
very nice PR as usual! LGTM functionally speaking

PR #364, opened 2018-11-09
Fixing crash after adding credential in MMM and discarding save password

deXol:
By the time the lambda is called the captured **pwdChangeJobs** pointer contains memory garbage. During _setCurrentJobError_ function it crashes.
In _runAndDequeueJobs()_ function **currentJobs** is set to the actual job, which is valid, when the lambda is called.

**Crash Repro**:
1. Go to Credentials
2. Enter Credentials Management Mode with Mooltipass device
3. Add a new credential with "Add or update a credential" section
4. Click Save all changes
5. Mooltipass device asks "Change Password For" if the user discards it or timeouts, then the daemon crashes, the user cannot use MC, until restarting the daemon and the added credential is saved with some invalid password (originaly it was "test" too):
6. kep

**Occurrence**: every time
**Test on**: Windows, Ubuntu

After this fix it does not crash.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -6805,11 +6805,11 @@ void MPDevice::setMMCredentials(const QJsonArray &creds, bool noDelete,
 
                     pwdChangeJobs->append(new MPCommandJob(this, MPCmd::SET_LOGIN,
                                                   ldata,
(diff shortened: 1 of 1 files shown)


limpkin:
Nice find!

raoulh:
thanks

Issue #368, opened 2018-11-14
System notification next steps


limpkin:
Current notifications implemented are:
- simple notification when a credential storage is triggered ("Credentials detected!")
- text input when an empty login is detected for storage
- subdomain choice when enabled in moolticute

However, status notifications currently implemented on chrome / FF only aren't migrated (yet) to moolticute: when a credential recall/storage is needed, a notification letting the user know the device is locked / not connected (https://github.com/mooltipass/extension/blob/master/background/event.js#L290)

Obviously we wouldn't want to spam the user with notifications that the device is locked... so I suggest to show this kind of notification **only once** for a given device status and re-enable status notification once this status has changed (device has been unlocked for example)

How does moolticute know when to display this notification?

I suggest the following:
- whenever the extension needs to do a storage/recall, as it already knows the device status (reported by moolticute automatically), it could send a "show_status_notification_warning" message to moolticute, which would then show the correct simple notification message based on the current device status (please connect the device, please unlock the device....)

@deXol is planning to try QSystemTrayIcon notification for simple notifications on win7 and win8

Other pending actions:
- not launch snoretoast at first start on win 7
- use qdialog for empty login / subdomain for win < 10

limpkin:
@raoulh : is this way to do ok with you?

raoulh:
I have no objection

limpkin:
implemented.

PR #369, opened 2018-11-14
#368 Sending simple system notification with systray bellow Win10

deXol:
WinRT is not available on win7, so if the client wants to run SnoreToast it is getting an "_api-ms-win-core-winrt-l1-1-0.dll is missing_" error.

So on win7/win8 I implemented to send the simple notification through systray:
win7systraynotification

Also only checking if snoretoast is installed on win10.


limpkin:
very nice :)

raoulh:
thanks

PR #371, opened 2018-11-17
#368 Displaying notification on show_status_notification_warning mess…

deXol:
…age, depending on device state

Same messages used like on Extension side (https://github.com/mooltipass/extension/blob/master/background/event.js#L290), but conditions are fixed.

**Example**:
devicestatusnotification


limpkin:
extremely nice! I can't wait to get rid of notifications in the extension

Issue #370, opened 2018-11-16
Feature request: in MMM, a button to only display favorites


limpkin:
maybe add a small one on the left of the search field?

![image](https://user-images.githubusercontent.com/6170291/48632775-087ae880-e9c2-11e8-9cb2-9cdebfba7cb6.png)

limpkin:
implemented.

PR #374, opened 2018-11-20
Adding Favorite Filter to MMM

deXol:
I added a star icon to Credential management for filtering favorites credentials:
**Unfiltered**:
kep
Filtered:
kep

After filtering by favorites, name filter works too.


deXol:
Implemented #370

Issue #345, opened 2018-09-26
Add moolticute install folder to windows path


limpkin:
... in order to use mc-cli from the command line.

raoulh:
nope a path is a path ;)

limpkin:
so we're gently scr*wed?

raoulh:
I have a lot of other programs on windows where you need to first cd to_the_path to execute commands... That's windows...

Also it's never a good idea to change the user PATH variable on windows. It has drawbacks and can break things...

Something that could be done maybe, is installing a .bat file and a shortcut in start menu to start a cmd.exe that has the PATH changed with inclusion of moolticute folder.

limpkin:
or could we put mc-cli into a separate folder and add that folder to the PATH?

raoulh:
Yes we could do that too

limpkin:
seems a nice task for @asashnov then :)

limpkin:
though to add to the windows path we need admin rights, right @raoulh ?

raoulh:
I'm not sure, you can also set env var for the currrent user... so maybe not

limpkin:
what's the current status of that one @deXol ? why was it reverted?

deXol:
> 
> 
> what's the current status of that one @deXol ? why was it reverted?

hmm I think this was by mistake, the fix was on my master branch, and was reverted on a develop branch after a sync.
Will put it on a develop branch, and create a PR.

limpkin:
right!
IIRC we still need to modify the installation process to put mc-cli inside that created forlder, right @raoulh ?

raoulh:
Yes, I'll do that before the next prerelease

limpkin:
gentle reminder for @raoulh

PR #377, opened 2018-11-25
#345 Adding cli to PATH in Inno Setup

deXol:
It is added according to Inno Setup documentation:
http://www.jrsoftware.org/isfaq.php#env
http://www.jrsoftware.org/ishelp/index.php?topic=registrysection


limpkin:
@raoulh I don't remember: did you implement support for this on your side?

Issue #272, opened 2018-05-01
Inability to modify "Service or Website" from Moolticute "Credentials Management" mode


kbembedded:
**Issue:**
While rare, I sometimes find myself in a situation where I need to rename the Service/Website of an entry in my mooltipass.  I recently moved over to moolticute and found that the credentials management mode of the application is unable to rename the Service/Website.

Using v0.12.11 on both a Windows host and Debian host.

**Workarounds:**
Switch to using the old Mooltipass Chrome app to rename the required Service/Website names.
Create a new entry with the proper URL and delete the old and incorrect one.

alanrick:
Just noticed the same today.  E.g. remove the http:// prefix from some entries to improve sort order.

AtosNicoS:
I also noticed this bug today. It is really annoying and a big regression in the user experience. It would be nice if you can give the issue more focus, its been opened for over half a year now.

limpkin:
Hello everyone,

This particular issue hasn't been forgotten... We have focused our efforts
on things that do not have workarounds.

limpkin:
implemented.

PR #381, opened 2018-12-02
#272 Develop272 edit service name in mmm

deXol:
I merged @limpkin's  backend solution, and finished GUI side.

Added an edit icon next to the service name with tooltip:
kep

If the user enters already existing service name, then unable to save it:
kep

They can save it with valid service name, and after saving the credential tree view is ordered:
kep
kep


limpkin:
looks good to me @raoulh : you can merge and then merge to master if you're ok with it

deXol:
Travis failed on Unix, because I used operator[] overload for QJsonValue, which was introduced in Qt 5.10., but the Travis build is using Qt 5.9.6.
http://doc.qt.io/qt-5/qjsonvalue.html#operator-5b-5d
Fixed it.

Issue #247, opened 2018-03-16
moolticute GUI forgets does not read 'device settings' after reboot


randomisresistance:
The moolticute GUI seems to fail to read the 'device settings' from the mooltipass device properly. It just displays the default settings. The actual settings on the device seem not to be affected by this. 

![image](https://user-images.githubusercontent.com/8646929/37516718-5bb2f37c-290f-11e8-93c5-952a390b70c7.png)

limpkin:
thanks for the bug report!
out of curiosity, are the udev rules correctly installed?

randomisresistance:
Hi limpkin,

sorry for the delayed answer. I used the Mooltiapp before and therefore have the udev rules deployed for a while now. But I double checked and everything is set up correctly.

I also have to add that the problem disappears when I open up the Moltucute app and then restart the daemon via the tray icon while the GUI is open. So this seems to be a smaller glitch.

Hope this helps!

limpkin:
I just had the very same bug on a slow windows machine.
Is it possible it is related to #246 ?
tagging @raoulh

limpkin:
pinging @raoulh

limpkin:
I recently got this issue... could you keep an eye for it @deXol ?

limpkin:
solved with latest commit :)

PR #384, opened 2018-12-10
#247 Develop247 device settings read fix

deXol:
If autolaunch was enabled, then after OS start, the settings from the device were not loaded.
This was because in that case mainwindow wasn't created during startup, so when the daemon sent the parameter loaded signals, then the mainwindow could not process them.

### DIFF
--- a/src/AppGui.cpp
+++ b/src/AppGui.cpp
@@ -187,8 +187,7 @@ bool AppGui::initialize()
     dbMasterController = new DbMasterController(this);
     dbMasterController->setWSClient(wsClient);
 
-    if (!autoLaunched)
-        mainWindowShow();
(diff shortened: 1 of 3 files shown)


limpkin:
woohoo! very nice

raoulh:
Does it work on macOS? At startup on macOS it crashed when creating the main window too early with start at boot is enabled.

deXol:
> Does it work on macOS? At startup on macOS it crashed when creating the main window too early with start at boot is enabled.

Checked on Mac (only debug version, was not able to build release), it does not crash, works as before, but with the correct settings read.

limpkin:
i guess we can merge and beta release and double check?

raoulh:
Ok, let's try with a test release build

Issue #383, opened 2018-12-09
Adopting new USB HID protocol in MP Device


deXol:
### New protocol implementation on daemon side:
The idea is to create an **IMessageProtocol interface**, where we add generic Message functions as pure virtual.
Then specializing the IMessageProtocol interface to **MessageProtocolMini** and **MessageProtocolBLE** classes.
Adding this IMessageProtocol pointer to MPDevice, and when creating an object of MPDevice we initializing this with the correct protocol class:
- We add `IMessageProtocol *pMesProt` to MPDevice.
![kep](https://user-images.githubusercontent.com/11043249/49698746-340b8000-fbc8-11e8-8916-855ec7f2928c.png)

For this we need to **generalize the message related operations in MPDevice**.
This is basicly implementing the MessageProtocol with Abstract Factory design pattern:
The solution is easily scalable and maintainable, so if we have a new protocol it just need to implement the IMessageProtocol interface, and we can use it too.

**Example**:
Current solution for getting command id from message:
`data.at(MP_CMD_FIELD_INDEX)`
This is working fine with the current protocol, but we can generalize it:
`pMesProt->getCommand(data)`

And we have the implementation in MessageProtocolMini (current) and MessageProtocolBLE (new) and will call the correct one which is decided in runtime, depending on which device is connected.

**Another example:**
_Current_:
`set_flashMbSize((quint8)data.at(2));`
_New_:
`set_flashMbSize(pMesProt->getFirstPayloadByte(data))`

### With this solution it required to change the following parts of MPDevice:
1. **_MPDevice::sendData_: Sending messages to Mooltipass**
	- We can create the packages from the parameters of sendData.
	- The current implementation creates an **MPCommand** and fills its attributes.
	data attributes contains the message's payload:
		`cmd.data.append(data.size());`
		`cmd.data.append(c);`
		`cmd.data.append(data);`
	Instead of this we should use a **common getPackets function** to construct the packets. (Just like in _mooltipass_hid_device.py get_packets_from_message_)
	- Another change would be to use a vector for MPCommand's data, because with the new protocol we can **send multiple packets** in a message, and for the old protocol just using the first element of the vector.

2. **_MPDevice::newDataRead(const QByteArray &data):_** **Receiving message from Mooltipass**
	- There are a few parts, where it required to use the generalized solution.
		- E.g.: Checking command, Check if need to resend message
	- It is possible we need to change the processing of response with multiple packets.
[...]

PR #393, opened 2019-01-02
Testing&Review #383 MPDevice message protocol generalization

deXol:
PR contains:
- message protocol generalization in MPDevice
- old protocol should work exactly like before
- added device detection for BLE on Windows, Linux and Mac
     - Currently the message protocol for BLE is unimplemented, but it is detected and the correct protocol class is initialized, when BLE is connected

Branched from **develop383ImplementNewProtocolSupport** for testing and reviewing. I'm continuing to implement a message protocol for BLE on develop383ImplementNewProtocolSupport branch.


limpkin:
@raoulh : really looking forward to your review on that one :)

src/MPDevice.h
```diff
@@ -96,7 +98,7 @@ class MPDevice: public QObject
     QT_WRITABLE_PROPERTY(qint64, uid, -1)
 
 public:
-    MPDevice(QObject *parent);
+    MPDevice(QObject *parent, bool isBLE = false);
```
raoulh: I don't like so much this boolean that is passed all over constuctor... There is already a `bool isMiniFlag = false;` for the variant of Mooltipass (first edition the big one), the Mini, and now the BLE. It would be better to have some kind of enum somewhere to keep the device type. Booleans spreads all over the code are not so good for maintenance...

deXol:
Thank you @raoulh 
Fixed the issues.

src/MPDevice.h
deXol: Created an enum for device type, but needed to move the message protocol init from base class constructor.

PR #404, opened 2019-02-01
Testing&Review #383 message protocol generalization

deXol:
### Changeset:
- **Added BLE Dev tab**
     - Features: Upload Bundle File, get platform informations, flash main, aux MCU
     - Tested UI layout on Windows, Linux, Mac
     - Available with CTRL+SHIFT+F3
     - kep

- **Implemented missing BLE Dev Message protocol functions**
     - MessageProtocolBLE class
- **Added generic commandIds for command mapping**
     - On BLE commandIds are different from mini/classic, so created a mapping for them.
- **Waiting every response packet for BLE to process the payload in a whole**
     - For BLE it is waiting while every packet is received, because the response has information about packet id and full packet number.
     - For the first packet the entire packet (not just the payload) is added to the response QByteArray for backward compatibility reasons with Mini and Classic, hence we can use generic message protocol interfaces for command implementations in MPDevice.
     - Tested with **0x8008**: Get 32 Samples of Accelerometer Data and **0x800A**: Get Platform Info
- **Added MPDeviceBleImpl class**
     - Implementations of only BLE related commands and helper functions.
     - Inharitance was not possible, because of the current class hierarchy of MPDevice (emul, mac, win, linux), therefor I used a Bridge like pattern for these implemenations
     - Currently it contains every BLE Tab related command implementations
     - This was required because MPDevice is already huge and I did not want to grow it further
- **Adopting MPDevice_emul with messageprotocol for mini device**
- **Disabled retry logic for BLE**


src/MPDeviceBleImpl.cpp
```diff
@@ -0,0 +1,216 @@
...
+        }
+        else
+        {
+            /**
+              * Main MCU flash does not cause a reconnect of the device,
+              * so daemon keeps sending messages to the device, which
+              * is causing failure in the communication, this is why
+              * this sleep is added.
+              */
+            qDebug() << "Waiting for device to start after Main MCU flash";
```
limpkin: 👍

src/MPDeviceBleImpl.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    if (0x01 == bleProt->getFirstPayloadByte(data))
+    {
+        qDebug() << "Erase done in: " << timer->nsecsElapsed() / 1000000 << " ms";
+        delete timer;
+        sendBundleToDevice(filePath, jobs, cbProgress);
+    }
+    else
+    {
+        //Erase is not done yet.
```
limpkin: actually, dataflash isn't ready, but nvm...

src/MPDevice_mac.cpp
```diff
@@ -83,8 +89,7 @@ QList<MPPlatformDef> MPDevice_mac::enumerateDevices()
 
 void MPDevice_mac::platformWrite(const QByteArray &data)
 {
-    //Do the write operation in a thread to avoid blocking
-    QtConcurrent::run([=]()
+    QtConcurrent::run(usbWriteThreadPool, [=]()
```
limpkin: 👍

src/MPDeviceBleImpl.cpp
```diff
@@ -0,0 +1,216 @@
...
+        else
+        {
+            /**
+              * Main MCU flash does not cause a reconnect of the device,
+              * so daemon keeps sending messages to the device, which
+              * is causing failure in the communication, this is why
+              * this sleep is added.
+              */
+            qDebug() << "Waiting for device to start after Main MCU flash";
+            QThread::sleep(5);
```
raoulh: This blocks the UI....

You should append a new TimerJob, which will delay the job queue

src/MPDevice_mac.cpp
```diff
@@ -47,6 +47,12 @@ MPDevice_mac::MPDevice_mac(QObject *parent, const MPPlatformDef &platformDef):
         deviceType = DeviceType::BLE;
     }
     setupMessageProtocol();
+    /**
+      * With only one thread for the threadpool
+      * the writes will keep the original order.
+      */
+    usbWriteThreadPool = new QThreadPool(this);
+    usbWriteThreadPool->setMaxThreadCount(1);
```
raoulh: Clever !

src/WSServerCon.cpp
```diff
@@ -684,6 +685,63 @@ void WSServerCon::processMessage(const QString &message)
             sendJsonMessage(oroot);
         });
     }
+    else
+    {
+        //Only ble related commands
+        MPDeviceBleImpl *bleImpl = mpdevice->ble();
```
raoulh: check the validity of `bleImpl`. If it's nullptr because the device is not a ble, it will crash

src/MPDevice.h.orig
```diff
@@ -0,0 +1,468 @@
+/******************************************************************************
```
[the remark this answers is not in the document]
deXol: ops I missed this one with merge tool

src/MPDeviceBleImpl.cpp
deXol: I tested this a little bit, and because it is on daemon's thread the UI was not blocked.
But I wasn't aware of TimerJob, it is a much clearer solution. And solves one more problem, because I can send it right after the flash command. (Do not have to wait until the flash command fails)

limpkin:
damn it... now we have conflicts :/

deXol:
> 
> 
> damn it... now we have conflicts :/

I will merge the master to this branch.

Issue #403, opened 2019-01-23
Bug: entering MMM and denying prompt


limpkin:
While running the 0.30.11-testing version, I came across some unexpected 
behavior. Here's how to reproduce it.

Open the app. Switch to the credentials tab and click the Enter 
Credentials Management Mode button. On the MPM, deny access to 
Management Mode, or just let it time out. A dialog box pops up that 
says, "Memory Management Error. An error occurred when trying to go into 
Memory Management Mode. Couldn't Load Database, Please Approve Prompt On 
Device" If you click on the dialog box's OK button the dialog goes away, 
but the app remains waiting for something to happen with the MPM -- it 
continues to show the rotating broken circle.

At this point, there's no way to confirm Memory Management Mode on the 
MPM since it is displaying the "home screen," so it looks like you're 
stuck. Luckily, there's a way out: switch to a different tab, then 
switch back.

FWIW, as I see it, the expected behavior when access is denied would be 
for the app's credentials panel to switch back to its initial display, 
noting with a message in the panel that access to the database was 
denied. I don't think a dialog box is a good idea in this case because 
the the user has actively denied access. If access fails because the 
request timed out on the MPM, a dialog seems okay -- but not strictly 
needed -- because it notes that the user's last request (i.e., to enter 
management mode) wasn't carried out. But in either case the error 
message should include something about having to try again. If the app 
can't tell the difference between a time-out and an active denial, then 
my preference would be to have no dialog box in either case and just 
note what happened with a message about having to try again on the app's 
credentials panel.

limpkin:
solved

PR #405, opened 2019-02-04
#403 Fix MMM denied stuck

deXol:
**Original issue:**
Enter Credentials Management Mode and deny it on the device. The app stucks at the waiting screen with a messagebox.
creddenystuck

**Fix:**
Instead of the waiting screen and the message box, displaying the original credentials with a message prompt, when enter MMM is denied. According to the expected behavior described in #403.
creddenyfix

The fix resolve the same issue on Files Management screen too.
I extended PromptWidget to support a simple information message with an OK button which hides the prompt.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1418,9 +1418,9 @@ void MainWindow::updateTabButtons()
 void MainWindow::memMgmtModeFailed(int errCode, QString errMsg)
 {
     Q_UNUSED(errCode)
-    QMessageBox::warning(this,
-                         tr("Memory Management Error"),
-                         tr("An error occured when trying to go into Memory Management mode.\n\n%1").arg(errMsg));
+    updatePage();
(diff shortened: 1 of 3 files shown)


raoulh:
On your gif video, before denying MMM, you had a previous prompt message showing the "no database  backup" text. After denying MMM, a new prompt message comes in, where you click ok. But then everything is gone, even the "no database backup" prompt. Is it a wanted feature?

@limpkin what do you think?

limpkin:
@raoulh : it's a side effect I can live with... this is just a gentle reminder anyway.

deXol:
I was thinking about a solution where I save the latest non-informative PromptMessage and load it back if the user click on OK for the current. On the other hand I do not know if it worth it if the user hasn't clicked on the prompt option by the time this error occurs.
Another possible issue if the user does not click on OK, then it will stay and it can mislead them.

limpkin:
@raoulh I'm fine with that... the bottom notification is considered as "low importance stuff"

Issue #410, opened 2019-02-14
Unicode BMP support for BLE


deXol:
**Implementing support for Unicode BMP on BLE**
With using the MessageProtocol class added in #383.

In message protocol I create two functions for QString to ByteArray and backward conversion.
For mini/classic it is using utf8 encoding to store and display strings (service, login, pwd... etc) on the device.
I am generalizing these conversions which are called in MPDevice and MPNode.
QChar is stored as Unicode BMP, so it is straightforward to get the Unicode BMP character code and add it to a ByteArray.

limpkin:
is it done that one? :)

deXol:
yes, it is done

PR #411, opened 2019-02-14
#410 Add unicode support for BLE and generic conversions to MessageProtocol

deXol:


src/MPDevice.cpp
```diff
@@ -5242,7 +5242,7 @@ bool MPDevice::readExportFile(const QByteArray &fileData, QString &errorString)
             else if  (encryptionMethod == "none")
             {
                 /* Legacy, not generated anymore */
-                return readExportPayload(QJsonDocument::fromJson(importFile.value("payload").toString().toUtf8()).array(), errorString);
+                return readExportPayload(QJsonDocument::fromJson(pMesProt->toByteArray(importFile.value("payload").toString())).array(), errorString);
```
limpkin: if I'm not mistaken, this is not needed here as you are reading a file generated by moolticute and not the device

deXol: right, fixed it.

limpkin:
LGTM @raoulh

Issue #402, opened 2019-01-18
Feature request: integreate with HIBP


limpkin:
in short, add an option in the settings tab to check if a password has been compromised every time a credential is stored or recalled.
On detection, trigger a notification. Technical details here: https://www.troyhunt.com/the-773-million-record-collection-1-data-reach/

limpkin:
must be made as an option!

PR #412, opened 2019-02-18
#402 Integrate hibp

deXol:
Added Have I Been Pwned password check for credential is stored or recalled.

hibpnotification

**Note:** It is set to false by default, need to enable in settings

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -182,6 +182,7 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
     ui->pushButtonIntegrity->setStyleSheet(CSS_BLUE_BUTTON);
     ui->btnPassGenerationProfiles->setStyleSheet(CSS_BLUE_BUTTON);
     ui->pushButtonSubDomain->setStyleSheet(CSS_BLUE_BUTTON);
+    ui->pushButtonHIBP->setStyleSheet(CSS_BLUE_BUTTON);
 
     // Don't show the "check for updates" button when built from git directly.
     ui->pushButtonCheckUpdate->setVisible(QStringLiteral(APP_VERSION) != "git");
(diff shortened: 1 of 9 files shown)


src/MainWindow.cpp
```diff
@@ -1740,3 +1760,28 @@ void MainWindow::on_pushButtonSubDomain_clicked()
...
+void MainWindow::on_pushButtonHIBP_clicked()
+{
+    QSettings s;
+
+    bool en = s.value("settings/enable_hibp_check", false).toBool();
+
+    int ret;
+    if (en)
+    {
+        ret = QMessageBox::question(this, "Moolticute", tr("Disable HIBP check?"));
```
raoulh: The name HIBP is not common, user will get confused, please use the full service name

src/MainWindow.cpp
```diff
@@ -1740,3 +1760,28 @@ void MainWindow::on_pushButtonSubDomain_clicked()
...
+    }
+    else
+    {
+        ret = QMessageBox::question(this, "Moolticute", tr("Enable HIBP check?"));
```
raoulh: Same here

src/MainWindow.ui
```diff
@@ -3028,6 +3028,43 @@ Hint: keep your mouse positioned over an option to get more details.</string>
               </item>
              </layout>
             </item>
+            <item>
+             <layout class="QHBoxLayout" name="horizontalLayout_46">
+              <item>
+               <widget class="QLabel" name="labelHIBPCheck">
+                <property name="text">
+                 <string>HIBP check: Enabled</string>
```
raoulh: Same here HIBP -> Have I Been Pwnd

Also it would be nice to have a clickable link below that goes to the https://haveibeenpwned.com website to let user see what this service is.

deXol:
Resolved the issues with the latest commit.

raoulh:
thanks!

Issue #415, opened 2019-02-28
Display correct device informations for BLE


deXol:
Display correct device informations for BLE in About tab:
![kep](https://user-images.githubusercontent.com/11043249/53595450-95c30100-3b9d-11e9-9718-45889c9c0762.png)

With using the new [0x0003: Get Platform Info command](https://mooltipass.github.io/minible/protocol) the serial number is available and also add Main and Aux MCU version for BLE.

limpkin:
why is this is an issue? :D

deXol:
Just created for tracking changes and this way I can commit to this ticket.

limpkin:
oh ok :)

PR #416, opened 2019-03-02
#415 Display device informations for BLE on About tab

deXol:
Rename the BLE Tab related platInfo functions to debug and added a new one for handling 0x0003 Get Platform Info command.
On GUI side, if BLE is detected Main and Aux MCU are also displayed:
kep


limpkin:
@raoulh can you review and merge?

PR #422, opened 2019-03-16
Implemented a new usb add&remove handle for exclusive device open crash fix on Windows

deXol:
### Issue
On Windows when device is added or removed checkUsbDevices() is called. When there is an **exclusively opened device** it does not find the device again, because cannot open it for checking the product and vendor id, hence the MPDevice_win::enumerateDevices() function returns with an empty device list, so it disconnects every device.
  The crash is caused by the delayed exitMemMgmtMode, when we are calling it the device has already disconnected because of the issue mentioned above.

Also added an option for exclusive device opening, because when we are only peaking the device's vendor and product id from attributes it does not require an exclusive opening.

### Linux, Mac
If exclusive device opening will implemented for Linux or Mac we also need to check this issue and use this new handle for them too, if it is necessarily.

### DIFF
--- a/src/MPDevice_win.cpp
+++ b/src/MPDevice_win.cpp
@@ -80,7 +80,7 @@ HANDLE MPDevice_win::openDevice(QString path, bool exlusive /* =false */)
                            nullptr,
                            OPEN_EXISTING,
                            FILE_FLAG_OVERLAPPED,
-                           0);
+                           nullptr);
 
     if (GetLastError() == ERROR_ACCESS_DENIED)
         h = CreateFileA(qToChar(path),
@@ -89,7 +89,7 @@ HANDLE MPDevice_win::openDevice(QString path, bool exlusive /* =false */)
                         nullptr,
                         OPEN_EXISTING,
                         FILE_FLAG_OVERLAPPED,
-                        0);
+                        nullptr);
 
     return h;
 }
--- a/src/MPDevice_win.cpp
+++ b/src/MPDevice_win.cpp
@@ -71,12 +71,13 @@ QString MPDevice_win::getLastError(DWORD err)
     return result;
 }
 
-HANDLE MPDevice_win::openDevice(QString path)
+HANDLE MPDevice_win::openDevice(QString path, bool exlusive /* =false */)
 {
+    DWORD sharedOptRW = exlusive ? 0 : FILE_SHARE_READ | FILE_SHARE_WRITE;
     HANDLE h = CreateFileA(qToChar(path),
                            GENERIC_WRITE | GENERIC_READ,
-                           0, //Open exclusively
-                           NULL,
+                           sharedOptRW,
+                           nullptr,
                            OPEN_EXISTING,
                            FILE_FLAG_OVERLAPPED,
                            0);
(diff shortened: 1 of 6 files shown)


src/MPDevice_win.cpp
```diff
@@ -71,31 +71,32 @@ QString MPDevice_win::getLastError(DWORD err)
...
 
-HANDLE MPDevice_win::openDevice(QString path)
+HANDLE MPDevice_win::openDevice(QString path, bool exlusive /* =false */)
 {
+    DWORD sharedOptRW = exlusive ? 0 : FILE_SHARE_READ | FILE_SHARE_WRITE;
     HANDLE h = CreateFileA(qToChar(path),
                            GENERIC_WRITE | GENERIC_READ,
-                           0, //Open exclusively
-                           NULL,
+                           sharedOptRW,
```
limpkin: same as before?

deXol: Yes, opening device exclusivly, except during enumerate, when it is peaking the pid/vid.

src/MPManager.cpp
```diff
@@ -39,8 +39,8 @@ bool MPManager::initialize()
     if (!AppDaemon::isEmulationMode())
     {
 #if defined(Q_OS_WIN)
-        connect(UsbMonitor_win::Instance(), SIGNAL(usbDeviceAdded()), this, SLOT(usbDeviceAdded()));
-        connect(UsbMonitor_win::Instance(), SIGNAL(usbDeviceRemoved()), this, SLOT(usbDeviceRemoved()));
+        connect(UsbMonitor_win::Instance(), SIGNAL(usbDeviceAdded(QString)), this, SLOT(usbDeviceAdded(QString)));
+        connect(UsbMonitor_win::Instance(), SIGNAL(usbDeviceRemoved(QString)), this, SLOT(usbDeviceRemoved(QString)));
```
raoulh: The signal/slots signature changed from usbDeviceAdded() to usbDeviceAdded(QString), this change was not reflected in UsbMonitor_mac/UsbMonitor_linux, and this will build but fail at runtime. You need to change the signals/slots in those class too.

deXol: This has been only implemented for Windows yet (because of the crash), Mac/Linux is still using the usbDeviceAdded() without parameter.

raoulh: Yes I know but you changed the signature of the slots in MPManager
https://github.com/mooltipass/moolticute/pull/422/files#diff-afdcd22bf9d85fd4e0aad3e2ee1eb243R58

So it will be changed for windows, but also for linux and macos

deXol: I kept the signal without the parameter too, and for mac/linux connecting that one: https://github.com/mooltipass/moolticute/blob/21d3bd8d6c174cf269ce499f69fcfaac102b7d77/src/MPManager.cpp#L50
I tested it on Linux and called the correct function, hence I thought qt connect can handle overloads, but I can change the name of the function with QString arg instead of the overload.

raoulh: Ah you are right, I misread, you kept the old slots... It's ok then

PR #426, opened 2019-03-26
Implement store/get credential support for BLE on daemon side

deXol:
### Changeset:

- Implement store/get credential support for BLE on daemon side
    - Currently not available from gui, but I left commented example usage in MPDevice.cpp:151
    - Store a credential with service name, login, description, third field and password (**0x0006: Store Credential**)
    - Get credential with service name and login (**0x0007: Get Credential**)
    - Created a get_message interface to get and return the credential for a request from GUI
- Eliminated message timeout for BLE:
    - Daemon is waiting until it does not receive the response for the message
    - Handled the **checkReturn** option for messages, which does not require response from device. (E.g.: 0x8007: Reboot main MCU to Bootloader or 0x8009: Flash Aux MCU with Bundle Contents)
    - **Known issue**: After Aux MCU Flash the usb is reconnecting and daemon does not receive an answer for the first message. Need to check it on fw side.
- Fixed handle for PLEASE_RETRY message:
    - Before sending the message again with a delay, needed to flip the message flip bit.
- Fixed some BLE message protocol functions
    - Found issues during usage in getMessageSize, getFullPayload, getPayloadBytes and Unicode conversions

### DIFF
--- a/daemon.pro
+++ b/daemon.pro
@@ -98,7 +98,8 @@ HEADERS  += \
     src/MessageProtocol/MessageProtocolMini.h \
     src/MessageProtocol/MessageProtocolBLE.h \
     src/MPDeviceBleImpl.h \
-    src/HaveIBeenPwned.h
+    src/HaveIBeenPwned.h \
+    src/BleCommon.h
 
 DISTFILES += \
     src/http-parser/CONTRIBUTIONS \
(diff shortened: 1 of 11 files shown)


limpkin:
very nice job! that was fun to work on

PR #428, opened 2019-03-31
Implemented linux hidraw add remove handle

deXol:
Same handle, which was implemented for Windows in #422.
Handle adding/removing devices, instead enumerating devices every time.


deXol:
Implemented the same usb handle for Mac too.

PR #434, opened 2019-04-09
Sending a notification when exclusive grab failed on Linux

deXol:
If exclusive grab failed on usb connection, then sending the following notification:
kép

### DIFF
--- a/src/MPDevice_linux.cpp
+++ b/src/MPDevice_linux.cpp
@@ -52,6 +52,12 @@ MPDevice_linux::MPDevice_linux(QObject *parent, const MPPlatformDef &platformDef
         if (INVALID_VALUE == grabbed)
         {
             qWarning() << "Exclusive device grab wasn't successful: " << strerror(errno);
+            //Need a delay for setting MPDevice to WSServer
+            QTimer::singleShot(100,
+                [this]()
+                {
+                    sendNotification(tr("Exclusive Grab Failure"), tr("Moolticute couldn't open the USB port exclusively"));
+                });
         }
         sockNotifRead = new QSocketNotifier(devfd, QSocketNotifier::Read);
         sockNotifRead->setEnabled(true);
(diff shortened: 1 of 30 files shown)


src/MPDevice_linux.cpp
```diff
@@ -52,6 +52,12 @@ MPDevice_linux::MPDevice_linux(QObject *parent, const MPPlatformDef &platformDef
         if (INVALID_VALUE == grabbed)
         {
             qWarning() << "Exclusive device grab wasn't successful: " << strerror(errno);
+            //Need a delay for setting MPDevice to WSServer
+            QTimer::singleShot(100,
+                [this]()
+                {
+                    sendNotification(tr("Exclusive Grab Failure"), tr("Moolticute couldn't open the USB port exclusively"));
```
raoulh: The translation will not work like that. The QTranslator is installed in the gui process, not in the daemon, and thus when the strings will be passed like that they will remain always in english

deXol: Thanks I was not aware of this, so I moved the translation texts to GUI.

deXol:
Hello @raoulh,
I fixed the review issue, but I think this PR should not be merged yet.
I've checked the exclusive grab on a native linux (Ubuntu 18.04), and for me it is failing every time, just like on the virtualbox image (Ubuntu 16.04).
@limpkin mentioned you were able to acquire the exclusive grab on linux.
Were you using `ioctl(devfd, EVIOCGRAB, ExclusiveAccess::GRAB);` (ExclusiveAccess::GRAB == 1) too or implemented in an other way?
For me EVIOCGRAB fails every time with **22 errno** (Invalid argument)

raoulh:
I did some test and read the kernel source... Actually the `EVIOCGRAB` `ioctl` is not implement for hidraw, but only for evdev. It seems we cannot do exclusive grab on linux with hidraw...

https://elixir.bootlin.com/linux/v5.0.7/source/drivers/hid/hidraw.c#L373

https://elixir.bootlin.com/linux/v5.0.7/source/drivers/input/evdev.c#L1126

deXol:
Hmm thats too bad, I am closing this PR and removing the EVIOCGRAB in an other one.

PR #440, opened 2019-04-29
Implemented entering MMM on BLE

deXol:
Changeset:
- Created device specialized functions for MPNode, these are implemented in MPNodeMini and MPNodeBLE
- Eliminated most magic numbers from MPNode
- Created a factory method for the MessageProtocol, it is creating a MPNode object depending on the device is a Mini or a BLE.
- This factory method is called during creation of MPNode in MPDevice
- Moved MPNode and the derived classes to a Mooltipass folder _(maybe should move every MP* for more transparent folder structure)_
- Updated **0x0004: Set Current Date** message for BLE according new payload structure
- Mapped **0x0011: Get Status** to the new message for BLE
- Added a new interface for convertDate according to the device in the corresponding MessageProtocol
- Added comments and todos, where need to remove isBLE conditions, when the correct messages are implemented for BLE too.


src/Mooltipass/MPNodeMini.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+MPNodeMini::MPNodeMini(QObject *parent, const QByteArray &nodeAddress, const quint32 virt_addr)
+    : MPNode(parent, nodeAddress, virt_addr)
+{
+}
+
+bool MPNodeMini::isDataLengthValid() const
+{
+    return data.size() == MP_NODE_SIZE;
```
limpkin: are you sure about this one? node size differs depending on node type

src/MPDevice.cpp
```diff
@@ -1308,7 +1361,14 @@ void MPDevice::startMemMgmtMode(bool wantData,
     jobs->append(startMmmJob);
 
     /* Load flash contents the usual way */
-    memMgmtModeReadFlash(jobs, false, cbProgress, !wantData, wantData, true);
+    if (isBLE())
+    {
+        memMgmtModeReadFlashBLE(jobs, false, cbProgress, !wantData);
```
limpkin: any reason to remove the extra parameters for that function clone? they are needed in the previous gen...

src/MPDevice.cpp
```diff
@@ -1287,6 +1293,53 @@ void MPDevice::memMgmtModeReadFlash(AsyncJobs *jobs, bool fullScan,
     }
 }
 
+//TODO: Remove if favorites, cpz_ctr, ctrvalue and getData implemented for BLE too
+void MPDevice::memMgmtModeReadFlashBLE(AsyncJobs *jobs, bool fullScan, const MPDeviceProgressCb &cbProgress, bool getCreds)
+{
+    if (getCreds)
+    {
+        /* Get parent node start address */
+        jobs->append(new MPCommandJob(this, MPCmd::GET_STARTING_PARENT,
```
limpkin: to be fair it should be called get starting parents :)

src/Mooltipass/MPNode.cpp
```diff
@@ -25,26 +25,35 @@ QByteArray MPNode::EmptyAddress = QByteArray(2, 0);
 
 MPNode::MPNode(const QByteArray &d, QObject *parent, const QByteArray &nodeAddress, const quint32 virt_addr):
     QObject(parent),
-    data(std::move(d)),
-    address(std::move(nodeAddress)),
+    data(d),
+    address(nodeAddress),
```
limpkin: sure about the change? IIRC this is quite a hack to copy the data.....

src/Mooltipass/MPNodeBLE.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+bool MPNodeBLE::isValid() const
+{
+    if (NodeUnknown == getType())
+    {
+        return false;
+    }
+
+    return (data.size() == PARENT_NODE_LENGTH || data.size() == CHILD_NODE_LENGTH);
```
limpkin: ahaha I see that working given how I send data... but better look at the flags to know which data type it is: 
https://github.com/mooltipass/minible/blob/master/source_code/main_mcu/src/COMMS/comms_hid_msgs.c#L443

src/MPDevice.cpp
deXol: Yes, those message mapping havent been implemented for the BLE yet. Thats why a copied only getCredential to a new temporary function (**memMgmtModeReadFlashBLE**). After the commands are mapped it will use `memMgmtModeReadFlash` too just like the mini and I will remove `memMgmtModeReadFlashBLE`.

src/Mooltipass/MPNodeBLE.cpp
deXol: I have checked this from the database description: https://mooltipass.github.io/minible/database_model
According that parent nodes are 264B and Child nodes are 528B

src/Mooltipass/MPNode.cpp
deXol: I removed std::move because we were copying it previously too.
For a const lvalue reference (const T&) std::move is only casting to const T&&, which is still calling the copy constructor.
https://stackoverflow.com/questions/38917130/does-stdmove-work-with-lvalue-references-how-does-stdmove-work-on-standard

src/MPDevice.cpp
deXol: Mentioned in the header: https://github.com/mooltipass/moolticute/pull/440/files#diff-8419d70e7413e49a93c9287a8134a9ecR319

src/MPDevice.cpp
deXol: Now yes, but later it will be deleted. 
And during implementing the commands I will copy the other jobs from `memMgmtModeReadFlash` to here for testing it. :)

src/Mooltipass/MPNode.cpp
deXol: I pushed a solution for this to move QByteArray whenever possible.
There are 3 scenario to create MPNode in MPDevice:
- **const T& case**: `MPNode *pnode = pMesProt->createMPNode(this, address);`
    - Here `address` is given from function parameter (const T&) and used later.
    - After moving here would still copy the QByteArray
- **Temporary QByteArray**: `temp_node = pMesProt->createMPNode(temp_node_pt->getNodeData(), this, temp_node_pt->getAddress(), temp_node_pt->getVirtualAddress());`
   - Because of the RVO the new function with universal reference (T&&) will be called by default for MPNode and it will move the QByteArray
- **QByteArray as lvalue and not used later**: `MPNode* importedNode = pMesProt->createMPNode(dataCore, this, serviceAddr, 0);`
   - If I move the lvalue the correct function is called and it is moving the QByteArray
   - Changed to: `MPNode* importedNode = pMesProt->createMPNode(qMove(dataCore), this, qMove(serviceAddr), 0);`

src/Mooltipass/MPNodeBLE.cpp
deXol: should I check the type here beside the size?

src/MPDevice.cpp
limpkin: ok!

src/MPDevice.cpp
limpkin: ok!

src/Mooltipass/MPNodeBLE.cpp
limpkin: you need to check the size for a given node type. if the node is a child, you need to make sure it is 528B for example

src/Mooltipass/MPNode.cpp
limpkin: calling @raoulh  on this one

src/Mooltipass/MPNodeBLE.cpp
deXol: added the check in the latest commit

src/Mooltipass/MPNode.cpp
raoulh: @deXol is right

PR #445, opened 2019-05-11
Implement BLE commands

deXol:
Following commands were implemented:
- 0x000A / get change numbers
- 0x000B / Get card CPZ
- 0x0005: Cancel Request
- 0x000E / reset unknown card
- 0x000F / get # of available users
     - Get number of available users was not implemented for mini too, so I added this to GUI on Advanced Tab and the user is able to get this information for BLE and Mini.
- 0x0010 / lock device
- 0x0011 / get status

### DIFF
--- a/src/MessageProtocol/MessageProtocolBLE.cpp
+++ b/src/MessageProtocol/MessageProtocolBLE.cpp
@@ -128,6 +128,14 @@ AsyncFuncDone MessageProtocolBLE::getDefaultFuncDone()
     };
 }
 
+AsyncFuncDone MessageProtocolBLE::getDefaultSizeCheckFuncDone()
+{
+    return [this](const QByteArray &data, bool &) -> bool
+    {
+        return getMessageSize(data) != 0;
+    };
+}
+
 QString MessageProtocolBLE::getDeviceName() const
 {
     return "BLE";
--- a/src/MessageProtocol/MessageProtocolBLE.cpp
+++ b/src/MessageProtocol/MessageProtocolBLE.cpp
@@ -93,6 +93,13 @@ quint32 MessageProtocolBLE::getSerialNumber(const QByteArray &data)
     return 0;
 }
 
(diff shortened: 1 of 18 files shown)


limpkin:
Hey there!

Awesome work!
Did you run/update the unit tests that can be found in the qt creator solution?

deXol:
> 
> 
> Hey there!
> 
> Awesome work!
> Did you run/update the unit tests that can be found in the qt creator solution?

I've just checked them, every Unit Test passed. however there are no tests related to this changeset. (Only for Models in GUI, Updater, FileCache and ParseDomain)

limpkin:
oh damn, I thought we had some... LGTM then :)

src/MessageProtocol/MessageProtocolBLE.cpp
```diff
@@ -156,6 +180,24 @@ void MessageProtocolBLE::setAckFlag(bool on)
     m_ackFlag = on ? ACK_FLAG_BIT : 0x00;
 }
 
+quint32 MessageProtocolBLE::convertToQuint32(const QByteArray &data)
+{
+    QVector<quint8> l(4, 0);
+    std::transform(std::begin(data), std::end(data), std::begin(l),
+                   [](char c) {return static_cast<quint8>(c);});
+    return convertToQuint32(l[0], l[1], l[2], l[3]);
```
raoulh: I don't like this...

Data are copied in a vector just to cast to quint8. Why not just doing:

```
return convertToQuint32(static_cast<quint8>(data.at(0)), 
                        static_cast<quint8>(data.at(1)),
                        static_cast<quint8>(data.at(2)),
                        static_cast<quint8>(data.at(3)));
```

deXol: data size can be under 4, thats why I did not implemented this way.

raoulh: Then a bunch of: ```if (data.size() < xxx)``` is way faster than allocating a QVector everytime...

PR #448, opened 2019-05-23
[BLE] Implement set user categories, add User Security Settings

deXol:
### Changeset:
- Implement 0x0012: Check Credential message
     - Added a check before adding a new credential from MC
- Implement Get/Set User Categories (0x0014/0x0015):
     - Only displaying on Credentials tab, when Advanced Mode is enabled and BLE device is connected:
      Screenshot from 2019-05-23 13-47-43

- Implement 0x0013: Get User Settings
     - Only displayed for BLE device
     - Read only, polling with status message
      bleUserSecuritySettings


limpkin:
this is really cool.

PR #451, opened 2019-06-09
Implement Device Settings refactor and add set/get device settings for BLE

deXol:
### The new design
kép
Created a new macro for Device Settings properties **QT_SETTINGS_PROPERTY**, which is storing the Param type too and sending it with `name##Changed signal`
Device setting properties which are stored on the device is moved from MPDevice on daemon side and from WSClient on GUI side to DeviceSettings.

With the new design when a new settings is added only need to change the following codes:
**Common:**
- Add new property to `DeviceSettings` if it is a common device settings, else to the corresponding device's class (`DeviceSettingsBLE/DeviceSettingsMini`)
     - `QT_SETTINGS_PROPERTY(int, keyboard_layout, 0, MPParams::KEYBOARD_LAYOUT_PARAM)`
- Add the parameter name mapping to `m_paramMap` in the corresponding device's class `fillParameterMapping()` function.
     - `{MPParams::KEYBOARD_LAYOUT_PARAM, "keyboard_layout"},`

**Daemon:**
- Implement loading parameter from device in **MPSettingsMini/MPSettingsBLE** `loadParameters()` function.
- **For BLE** the byte mapping of settings with parameter is stored in `m_bleByteMapping `map:
      - `m_bleByteMapping[MPParams::RANDOM_INIT_PIN_PARAM] = RANDOM_PIN_BYTE;`

**GUI:**
- New UI element is need to added to **MainWindow.ui** file.
- Mapping for parameter and UI element is added in void SettingsGuiHelper::setMainWindow(MainWindow *mw) function:
    - `{MPParams::KEYBOARD_LAYOUT_PARAM, ui->comboBoxLang}`
- Displaying device relevant settings can be implemented in `SettingsGuiMini::updateUI()` and `SettingsGuiBLE::updateUI()` function:
```
void SettingsGuiBLE::updateUI()
{
    ui->groupBox_BLESettings->show();
}
```

Every other signal&slot connections are connacted from code using metaobject property informations, so if a new property is added it will connect it dynamically.
**For Mini** the UI hasn't changed only the settings handling in the backend.
**For BLE** I have added `BLE Reserved` and `Prompt Animation` in a BLE Settings groupbox.
kép
The following settings parameters are connected based on BLE Message protocol page (https://mooltipass.github.io/minible/protocol):
- Reserved
- Random PIN
- User interaction timeout
- Animation during prompt

The rest of device settings haven't implemented on FW side, and will connacted later. But saving the mentioned settings above is working currently.

**Possible more improvement:**
- Move the UI widget from MainWindow.ui to a separate one.


raoulh:
Nice refactor, I wanted to do something like that for a long time

PR #473, opened 2019-08-27
Implemented save/modify/remove credentials in MMM for BLE

deXol:
For this the following commands were implemented and specialized for BLE:
- MPCmd::WRITE_FLASH_NODE
- MPCmd::SET_STARTING_PARENT
- MPCmd::GET_FREE_ADDRESSES
- MPCmd::SET_USER_CHANGE_NB
- MPCmd::SET_DATA_CHANGE_NB

So currently it is possible to add a new credentials in MMM, just like outside of MMM on Credentials tab.
Also the user can remove or modify existing credentials.

When password is changed (credential modified or new one added) it is possible to avoid prompting on BLE in the device settings.


limpkin:
awesome work!

src/MPDevice.cpp
```diff
@@ -4443,42 +4460,42 @@ void MPDevice::changeVirtualAddressesToFreeAddresses(void)
 {
     if (virtualStartNode != 0)
     {
-        qDebug() << "Setting start node to " << freeAddresses[virtualStartNode].toHex();
-        startNode = freeAddresses[virtualStartNode];
+        qDebug() << "Setting start node to " << getFreeAddress(virtualStartNode).toHex();
```
limpkin: how do you handle knowing if you're getting a free address for a parent node or child node, as they have different sizes?

src/MPDeviceBleImpl.cpp
```diff
@@ -480,6 +480,68 @@ void MPDeviceBleImpl::processDebugMsg(const QByteArray &data, bool &isDebugMsg)
...
+    }
+    auto addressPackage = addressFrom;
+    addressPackage.append(bleProt->toLittleEndianFromInt(m_parentNodeNeeded));
+    addressPackage.append(bleProt->toLittleEndianFromInt(m_childNodeNeeded));
+    jobs->append(new MPCommandJob(mpDev, MPCmd::GET_FREE_ADDRESSES,
+                                  addressPackage,
+                                  [this](const QByteArray &data, bool &) -> bool
+        {
+            const auto msgSize = bleProt->getMessageSize(data);
+            if ((m_parentNodeNeeded + m_childNodeNeeded)*2 != msgSize)
```
limpkin: maybe replace 2 with a define or something :D.
this is to tackle this: https://github.com/mooltipass/minible/blob/master/source_code/main_mcu/src/COMMS/comms_hid_msgs.c#L464 correct?

limpkin: maybe replace 2 with a define or something :D.
I think you still need to tackle this: https://github.com/mooltipass/minible/blob/master/source_code/main_mcu/src/COMMS/comms_hid_msgs.c#L464 , correct?

deXol: Switched to ADDRESS_LENGTH instead of 2.
Well I think it is handled with `(m_parentNodeNeeded + m_childNodeNeeded) * MPNode::ADDRESS_LENGTH != msgSize`, because in case of error the payload size will be 1, according to fw code, and then this will be true and the job returns with false.
But I can add an extra 1 check for msgSize too.

deXol:
### [WIP]
After discussing with @limpkin, currently implement handling more than 266 free address requested.

src/MPDevice.cpp
```diff
@@ -481,6 +481,13 @@ void MPDevice::sendDataDequeue()
         qDebug() << "Full packet#" << i++ << ": " << a;
 #endif
 
+        if (0 == i%5)
+        {
+            /* Added a 10 msec delay for long messages
+             * after every 5th packet to avoid packet drop
+             */
+            QThread::msleep(10);
```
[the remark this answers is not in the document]
raoulh: After discussing with limpkin about this issue, we need to have a correct fix for this. QThread::msleep is prohibited and a really bad solution.

Actually there is 2 solutions:
- implement throttling when sending packet here in `MPDevice::sendDataDequeue()`. Delay all paquet sending by at least a few ms between each. Problem is it will complexify even more this code (which already does a lot like packet retry, timeout, etc...)

- Implement a buffer in all MPDevice_xxx that waits for the data to be written by the system to the device. All OSs have this API available, and we use non blocking apis right now.

In Linux we do not wait for any data write events. The `QSocketNotifier` is created for `write()` but not used:
https://github.com/mooltipass/moolticute/blob/244c1796779762a42ed3d59ea6c53636c12ee071/src/MPDevice_linux.cpp#L55
Here we need to wait for that signal before writing the next data.

On Windows, this is the same, there is a `QWinOverlappedIoNotifier` but it's only checked for Read, write is being omited:
https://github.com/mooltipass/moolticute/blob/244c1796779762a42ed3d59ea6c53636c12ee071/src/MPDevice_win.cpp#L306

On Macos it's even simpler, because we use a second thread for writing the data as the API is blocking. We can just emit a signal after the write in the thread to notifiy the main thread to continue the write:
https://github.com/mooltipass/moolticute/blob/244c1796779762a42ed3d59ea6c53636c12ee071/src/MPDevice_mac.cpp#L108

src/MPDevice.cpp
```diff
@@ -5922,7 +5951,13 @@ void MPDevice::loadFreeAddresses(AsyncJobs *jobs, const QByteArray &addressFrom,
 {
     qDebug() << "Loading free addresses from address:" << addressFrom.toHex();
 
-    jobs->append(new MPCommandJob(this, MPCmd::GET_30_FREE_SLOTS,
+    if (isBLE())
+    {
+        bleImpl->getFreeAddressProvider().loadFreeAddresses(jobs, addressFrom, cbProgress);
```
limpkin: are you discarding discardFirstAddr? this is important as you'll get twice the same address in your buffer if not.

src/MessageProtocol/IMessageProtocol.h
```diff
@@ -153,6 +153,16 @@ class IMessageProtocol
         return littleEndian;
     }
 
+    QByteArray toLittleEndianFromInt32(quint32 num)
+    {
+        QByteArray littleEndian;
+        littleEndian.append(static_cast<char>(num&0xFF));
+        littleEndian.append(static_cast<char>((num&0xFF00)>>8));
+        littleEndian.append(static_cast<char>((num&0xFF0000)>>16));
+        littleEndian.append(static_cast<char>((num&0xFF000000)>>24));
```
limpkin: there's no MSb expansion here?
IIRC it's always better to put a '0' before the FF.

src/Mooltipass/MPBLEFreeAddressProvider.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            ba.append(receivedAddr.mid(pos, MPNode::ADDRESS_LENGTH));
+            pos += MPNode::ADDRESS_LENGTH;
+        }
+    }
+}
+
+void MPBLEFreeAddressProvider::loadRemainingFreeAddresses(AsyncJobs *jobs, const QByteArray &addressFrom, const MPDeviceProgressCb &cbProgress, bool isLastChild)
+{
+    auto freeAddrNum = MAX_FREE_ADDR_REQ - 1;
```
limpkin: why the -1? comments would be convenient

src/Mooltipass/MPBLEFreeAddressProvider.cpp
```diff
@@ -0,0 +1,144 @@
...
+    auto addressPackage = addressFrom;
+    FreeAddressInfo addressInfo;
+    addressInfo.parentNodeRequested = getNodeAskedNumber(MPNode::NodeParent, freeAddrNum);
+    addressInfo.childNodeRequested = getNodeAskedNumber(MPNode::NodeChild, freeAddrNum);
+    if (isLastChild)
```
limpkin: could you explain?

src/Mooltipass/MPBLEFreeAddressProvider.cpp
```diff
@@ -0,0 +1,144 @@
...
+    {
+        ++addressInfo.childNodeRequested;
+    }
+    else
+    {
+        ++addressInfo.parentNodeRequested;
+    }
+    addressPackage.append(bleProt->toLittleEndianFromInt(addressInfo.parentNodeRequested));
+    addressPackage.append(bleProt->toLittleEndianFromInt(addressInfo.childNodeRequested));
+    addressInfo.startingPosition = MPNode::ADDRESS_LENGTH;
```
limpkin: so with this variable you skip the first address already provided.
You really need to comment that code.

src/Mooltipass/MPBLEFreeAddressProvider.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+                int pos = addressInfo.startingPosition;
+
+                processReceivedAddrNumber(MPNode::NodeParent, receivedAddresses, pos);
+                processReceivedAddrNumber(MPNode::NodeChild, receivedAddresses, pos);
+
+                if (m_parentNodeNeeded + m_childNodeNeeded > 0)
+                {
+                    qDebug() << "Need more addresses: " << m_parentNodeNeeded + m_childNodeNeeded;
+                    const auto lastAddr = receivedAddresses.right(2);
```
limpkin: comments here

src/MPDevice.cpp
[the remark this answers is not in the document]
deXol: I implemented a writeBuffer solution in the latest commit for Windows.
I was wonder if race condition can be an issue here: 
https://github.com/mooltipass/moolticute/pull/473/files#diff-e9a41bdf95cd0a995d2c81602f7c5c00R461
https://github.com/mooltipass/moolticute/pull/473/files#diff-e9a41bdf95cd0a995d2c81602f7c5c00R5155
IIRC daemon's MPDevice related code is only running on one thread, so it should not be a problem.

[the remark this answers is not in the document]
raoulh: I checked your implementation. I'm not a fan, because you added a new queue in MPDevice. It would be better to put the queue in `MPDevice_win` and wait for the write signal there to continue sending data from the buffer like in `MPDevice_linux`:

https://github.com/mooltipass/moolticute/blob/173be4fc19c156cb2a67482dfb74662f2e84e3c9/src/MPDevice_linux.cpp#L102

It's better to have those mechanism in the platform code as it depends on the platform implementation and also `MPDevice` is already complicated enough to maintain another level of buffering there...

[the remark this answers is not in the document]
deXol: I can implement the queue in every platform's `MPDevice_xxx`, but I think the queue implementation is a generic one and only platform specific part is sending a request for the next data, when write is finished. (`emit platformWriteFinished()`)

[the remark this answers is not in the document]
deXol: > In Linux we do not wait for any data write events. The `QSocketNotifier` is created for `write()` but not used:
> 
> https://github.com/mooltipass/moolticute/blob/244c1796779762a42ed3d59ea6c53636c12ee071/src/MPDevice_linux.cpp#L55
> 
> Here we need to wait for that signal before writing the next data.
> 
I am checking Linux currently.  I connected `sockNotifWrite`'s activated signal for a slot, but the signal is never actually called after writes. I need further investigation regarding this.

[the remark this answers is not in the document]
deXol: I have moved the writeBuffer to `MPDevice_win`.
Also tested Linux and Mac there are no issues with writing longer messages. I tested with adding 20-30 credentials in MMM multiple times and have not experienced stuck.

deXol:
I investigated the possibility of **unit testing get free addresses class**.
I was not able to find a correct way of mocking in Qt Test Library, but I was thinking about using `MPDevice_emul` instead.
The first issue was because `MPBLEFreeAddressProvider` is heavily depending on `MPDevice` I need to include the entire daemon project. Including the project was not working (need more investigation), but I was able to include the files of it and compiled.
I made some dummy tests, and if I implement the BLE device's behavior for `MPCmd::GET_FREE_ADDRESSES` the unit tests will work, however as far I have checked `MPDevice_emul` is currently emulating the behaviour of the standard mini device, so it would be a little odd if every other message are implemented for that, except only GET_FREE_ADDRESSES. Therefor I think it would be nice to implement an emulation for BLE too and use that in the unit testing, if it worths the effort and there will be more unit testing for daemon in the future.

limpkin:
@raoulh let me know what you think. I personally think that this feature has been tested, please give it a thorough review and give me the green light so i can give it a final look & merge.
Regarding the buffer issue, it seems it is a windows-only one so a mitigation has only been implemented there. IIRC @deXol wasn't able to get the signals to work for linux as well.

raoulh:
Good also for me

limpkin:
woot!

PR #478, opened 2019-10-01
Persist user category names during export/import

deXol:
- Persist user category names and isBle information in the exported payload
- Loading user category names during import if available
- Changed magic number in import functions

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -4779,6 +4779,14 @@ QByteArray MPDevice::generateExportFileData(const QString &encryption)
     /* Mooltipass serial */
     exportTopArray.append(QJsonValue((qint64)get_serialNumber()));
 
+    if (isBLE())
+    {
+        /* isBle */
+        exportTopArray.append(QJsonValue{true});
+        /* user category names */
+        exportTopArray.append(bleImpl->getUserCategories());
+    }
+
     /* Generate file payload */
     QJsonDocument payloadDoc(exportTopArray);
     auto payload = payloadDoc.toJson();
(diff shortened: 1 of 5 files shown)


src/MPDevice.cpp
```diff
@@ -4905,16 +4909,20 @@ bool MPDevice::readExportPayload(QJsonArray dataArray, QString &errorString)
 {
     /** Mooltiapp / Chrome App save file **/
 
+    const auto dataSize = dataArray.size();
+    const bool isBleExport = EXPORT_BLE_DATA_SIZE == dataArray.size() && dataArray[EXPORT_IS_BLE].toBool();
+    const QString deviceVersion = dataArray[EXPORT_DEVICE_VERSION].toString();
     /* Checks */
-    if (!((dataArray[9].toString() == "mooltipass" && dataArray.size() == 10) || (dataArray[9].toString() == "moolticute" && dataArray.size() == 14)))
+    if (!((deviceVersion == "mooltipass" && dataSize == EXPORT_MP_DATA_SIZE)
```
limpkin: EXPORT_MP_DATA_SIZE might be confusing here... Maybe rename is as "number of fields for moolticute export"?

src/MPDevice.cpp
```diff
@@ -4905,16 +4909,20 @@ bool MPDevice::readExportPayload(QJsonArray dataArray, QString &errorString)
...
+          || (deviceVersion == "moolticute" && (dataSize == EXPORT_MC_DATA_SIZE || isBleExport))))
```
limpkin: same applies

src/MPDevice.cpp
```diff
@@ -4923,22 +4931,22 @@ bool MPDevice::readExportPayload(QJsonArray dataArray, QString &errorString)
...
         qDebug() << "Imported data change number: " << importedDataDbChangeNumber;
-        importedDbMiniSerialNumber = dataArray[13].toInt();
+        importedDbMiniSerialNumber = dataArray[EXPORT_DB_MINI_SERIAL_NUM].toInt();
         qDebug() << "Imported mini serial number: " << importedDbMiniSerialNumber;
     }
 
     /* Read CTR */
     importedCtrValue = QByteArray();
-    auto qjobject = dataArray[0].toObject();
+    auto qjobject = dataArray[EXPORT_CTR].toObject();
```
limpkin: maybe append with _INDEX this identifiers?

deXol: should I keep EXPORT_ too? because with both some of them will be really long

limpkin:
Why not.... We are in 2019 and have wide screens anyway no?

PR #482, opened 2019-10-21
[BLE] Implement favorite/category

deXol:
### Changeset:

- Implemented `GET_FAVORITE`, `SET_FAVORITE`, `GET_FAVORITES `commands
- Added DeviceDetector to GUI
    - There were several indepented UI classes where needed to get the device information, hence created a singleton class to handle it
- Added a folder icon with the credentials' category color next to the credential name
    - kép
- Handle simple mode:
    - Only favorite is displayed with the category's color
    - Category change is disabled
    - kép
- Implemented favorite logic for BLE:
    - The user can choose favorite 1-10 to the corresponding category 
    - Favorites are colored according to the category
    -  kép
 - Implemented category change in MMM

### Demo:
bleFavoriteCategoryDemo


raoulh:
Nice work

PR #488, opened 2019-11-07
Introduce support for devices connected over local sockets

anszom:
This PR introduces support for locally-connected devices in moolticuted, aimed to support the "emulator" build of mooltipass mini BLE. 
The code doesn't support other device models at the moment.


src/MPDevice.cpp
```diff
@@ -49,7 +49,8 @@ MPDevice::MPDevice(QObject *parent):
 
             if (isBLE())
             {
-                bleImpl->readUserSettings(pMesProt->getPayloadBytes(data, 2, 4));
+                if(pMesProt->getMessageSize(data) >= 4) // these settings are sent only when user is logged in
+                    bleImpl->readUserSettings(pMesProt->getPayloadBytes(data, 2, 4));
```
limpkin: Awesome catch...

moreover, they are now included in the get_status message... so you we may just remove that.... what do you think @deXol ?

deXol: this is the MPCmd::MOOLTIPASS_STATUS command's callback. In line 53 it is reading the user settings from get_status.

limpkin: oops, my bad!

limpkin:
merging #642 instead... thanks for the awesome work :)

PR #497, opened 2019-11-17
Make sure init messages are called first with a TimerJob

deXol:
Originally with the `QTimer::singleShot` delay, it was not guaranteed the init messages are called first, because during the 100 msec delay it was possible the GUI is requesting some data (e.g.: user category, language/layout related commands) and those were sent earlier causing a stuck in communication.
So I added a `TimerJob` first with a higher (300 msec) timeout, so there won't be any job added to the commandqueue before the init messages. However reset flip bit is written directly to the device without using the commandqueue, hence for reset flip bit a lower delay (150 msec) is still necessary.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -140,34 +140,41 @@ void MPDevice::setupMessageProtocol()
 
 void MPDevice::sendInitMessages()
 {
-    statusTimer->start(500);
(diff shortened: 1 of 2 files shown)


limpkin:
merging this as i'd like to test & deploy now @raoulh

raoulh:
good for me

Issue #505, opened 2019-11-20
Add timestamps to log


limpkin:
@deXol : to help the mysterious bug hunting, could you add timestamps in the log so we can correlate the moment we see the strange USB message and anything happening in moolticute?

limpkin:
implemented

PR #509, opened 2019-11-20
#505 Add timestamp to logs

deXol:
kép


limpkin:
Can you put it before all instead?

deXol:
> 
> Can you put it before all instead?

yes:
kép

limpkin:
Perfect!

Issue #486, opened 2019-11-07
Weird packet sent from moolticute


limpkin:
This is a very weird one...

moolticute log: 

DEBUG: :0 - "Received answer: GET_CUR_CARD_CPZ (0x000b)"  Full packet: "0c000b0xxxxxxxxxxxxx(hidden)00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
DEBUG: :0 - Card CPZ:  "<<hidden>>"
INFO: :0 - Finished loading card CPZ
INFO: :0 - "Loading device db change numbers"
DEBUG: :0 - Platform send command:  "GET_USER_CHANGE_NB (0x000a)"
DEBUG: :0 - Full packet# 0 :  "[0x04, 0x00, 0x0a, 0x00, 0x00, 0x00]"
DEBUG: :0 - JSON API recv: {
    "msg": "get_user_categories"

on the usb capture i can correctly see the get_card_cpz and its answer.... but I do not see the get_user_changer_nb packet at all. Instead, I see this one:

![image](https://user-images.githubusercontent.com/6170291/68416593-0db5e200-0195-11ea-9c61-9a2e7cac6de0.png)

which is an invalid packet....

deXol:
Is it happened on Windows?
It looks really weird, cannot really find any place where it was possible sent from MC.
What has happened after this message, is the GET_USER_CHANGE_NB messge not written at all?
And where you able to reproduce it since then?

limpkin:
windows indeed.
it is fairly reproduceable yes.... don't know about the get_user message

limpkin:
@deXol there's definitely something odd happening on the low level side of things.... how are buffers allocated? that looks like some buffer misuse.

This other packet is from a capture taken today:

![image](https://user-images.githubusercontent.com/6170291/69093725-513f0480-0a4f-11ea-8842-895065202ae8.png)

how come our websocket messages go to USB? this is high priority @deXol

raoulh:
That smells like a pointer to a wrong memory space, or a QByteArray that is uninitialized. I would advise @deXol to build moolticute using ASAN. To do so you can run qmake with:
```
qmake "CONFIG+=sanitizer sanitize_address" moolticute.pro
```
other sanitzers are also available:

- `sanitize_address`
- `sanitize_memory`
- `sanitize_thread`
- `sanitize_undefined`

deXol:
[...]

PR #515, opened 2019-11-24
#486 Fix winapi WriteFile buffer deallocation issue

deXol:
Analysed in https://github.com/mooltipass/moolticute/issues/486#issuecomment-557839802.

https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile#parameters

> lpBuffer
> 
> A pointer to the buffer containing the data to be written to the file or device.
> 
> **This buffer must remain valid for the duration of the write operation.** The caller must not use this buffer until the write operation is completed.

char* buffer was used from local variable `QByteArray ba`, but it is possible the write operation is not finished by the execution returns from `MPDevice_win::platformWriteToDevice` and in that case write operation is using a pointer to a deallocated buffer, which we can see in the #486 attached logs.

Copying the buffer into a new char* and deleting it when the write operation finished solves the issue.


raoulh:
Nice catch @deXol

PR #520, opened 2019-11-25
[BLE] Do not send get user categories request from GUI side in simple…

deXol:
… mode


limpkin:
keep in mind that the user may change modes when the device is connected to moolticute though!

deXol:
> keep in mind that the user may change modes when the device is connected to moolticute though!

Handled here:
https://github.com/mooltipass/moolticute/blob/master/src/CredentialsManagement.cpp#L196
https://github.com/mooltipass/moolticute/blob/master/src/CredentialsManagement.cpp#L1015

Issue #522, opened 2019-11-26
mini BLE: Device memory not displayed


dehne:
### Expected behavior
Display size of device memory on About tab after successful BLE connection.
### Actual behavior
Other device info is correctly displayed, but device memory displays "None".
### Step by step guide to reproduce the problem
1. Start from a clean slate: Unpair the mini BLE and the Windows 10 computer using the Windows "Bluetooth & other devices" Settings panel. Disable Bluetooth on the mini BLE. Quit Moolticute using the tray icon's Quit. Power off the mini BLE using its debug menu.
1. Reconnect the mini BLE via Bluetooth: Start Moolticute. Power up the mini BLE. Open Windows 10 Settings to "Bluetooth & other devices". Choose Add Bluetooth or other device. Unlock the mini BLE and navigate to the Bluetooth menu. Enable Bluetooth. Click "Add Bluetooth or other device" on the Windows settings panel. Watch Windows find the mini BLE. Once found, click on it in the settings panel to tell Windows to connect to it.
1. Once Windows connects, switch to the About tab in Moolticute. Observe that the device memory is listed as "None" but that the other device info is displayed.
![Moolticute001](https://user-images.githubusercontent.com/3382135/69655135-cb145680-102a-11ea-801e-c7174549fb66.png)

### Moolticute Version
42.13-testing
### Operating System
Windows 10
### Mooltipass Extension
N/A
### Mooltipass Device
Mooltipass Mini BLE

PR #570, opened 2020-01-01
[BLE] #522 Send device related informations to GUI on flashMbSize change too

deXol:
On startup device memory is not received from the device when sendVersion is called.
Calling sendVersion when flashMbSize is set fixes the issue.


deXol:
Fix for #522

Issue #526, opened 2019-11-26
Moolticute support for 64 characters


ThunderKing54:
8<------------------------ BUG REPORT -----------------------------------------
### Expected behavior
Moolticute should accept 64 characters for the Mooltipass mini BLE
### Actual behavior
Moolticute doesn't accept more than 31 characters  for the Mooltipass mini BLE
### Step by step guide to reproduce the problem
Add credentials with Moolticute:
1. go to credentials tab in Moolticute
2. add new Service or Site name
3. add username(Login)
4. try to add a password longer than 31 characters. After 31 characters the password TextBox won't accept more characters.

### Moolticute Version
v0.42.14-testing

### Operating System
Windows 10 Pro Build 18362
  
Other operating systems are not supported.

### Mooltipass Extension
- If you're creating an issue to report a website incompatibility, please use the "Report incompatibilities with this website" item in the extension menu.
- If you're creating an issue to report a problem with our extension, please create another issue here: 
https://github.com/mooltipass/extension/issues

### Mooltipass Device
- The Mooltipass Mini BLE

8<------------------------ END BUG REPORT -------------------------------------

PR #573, opened 2020-01-04
[BLE] Enable MMM password inputs to accept 64 characters

deXol:
PasswordLineEdits was limited to 31 characters, set this to 64 chars, when BLE is connected.
Also enabled to generate 64 long passwords.


deXol:
fix for #526

Issue #574, opened 2020-01-04
Feature to be implemented in MMM: key after login & password typing


limpkin:
see 260 to 263 in child node in https://mooltipass.github.io/minible/database_model

deXol:
@limpkin If I understand right I should add this one to the UI too and the users will be able to select key after login/password for each credential.

limpkin:
Yes :), maybe only if advanced mode is enabled?

deXol:
@limpkin I think it would be good when advanced mode is enabled.
What should be the default value for these when the user add a new credential?
And there is a global setting for key after login/password in Device settings:
![kép](https://user-images.githubusercontent.com/11043249/72208945-5a75ac00-34a9-11ea-9b27-29d54b5aa12c.png)
However if the keys are stored for every credentials what are these settings will used for?

limpkin:
By default it should be None, in that case the general setting takes over (maybe put this in the tool tip?)
If something is specified on the credential, then it takes over the global setting :)

PR #588, opened 2020-01-16
[BLE] Add Key After Login/Password option to MMM

deXol:
Implemented https://github.com/mooltipass/moolticute/issues/574.
Screen Shot 2020-01-16 at 21 37 42


limpkin:
very nice work!

Issue #586, opened 2020-01-15
WIP - tasks for database model change


limpkin:
Tasks given the move to more credential start adresses and less data start adresses
- HID_CMD_SET_CRED_ST_PARENT (0x0105) now has the same format as HID_CMD_SET_DATA_ST_PARENT to specify the credential type id

limpkin:
there are now 10 credential start addresses and 7 data start addresses (very important for backup file handling)

limpkin:
impacted commands:
- 0x0100: Get Start Nodes Addresses
- 0x0105: Set Credential Start Parent
- 0x0107: Set Start Nodes Addresses

PR #603, opened 2020-02-19
#586 Handle BLE database model change

deXol:
WIP, for review
Added support for webauthn credentials too and be able to handle further linked lists too.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -6114,7 +6114,7 @@ bool MPDevice::finishImportLoginNodes(QString &stringError, Common::AddressType 
             /* Marked for deletion? */
             if (!curNode->getMergeTagged())
             {
-                removeChildFromDB(nodeItem, curNode, true, true);
+                removeChildFromDB(nodeItem, curNode, true, true, addrType);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/MPDevice.cpp
```diff
@@ -1857,7 +1909,7 @@ bool MPDevice::tagPointedNodes(bool tagCredentials, bool tagData, bool repairAll
         }
     }
 
-    if (tagData)
+    if (tagData && isCred)
```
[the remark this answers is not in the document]
deXol: Yes, but it will be processed with the general credentials like for the mini.
I can add the comment, but we only need to modify this part if there will seperate data nodes for webauthn too.

src/MPDevice.h
```diff
@@ -393,7 +393,7 @@ private slots:
     quint32 credentialsDbChangeNumberClone;
     quint32 dataDbChangeNumberClone;
     QByteArray ctrValueClone;
-    QVector<QByteArray> startNodeClone = {0};
```
limpkin: oh i actually wondered about that. why the two elements though?

limpkin:
PR tested on mini & mini ble:
- from empty to "full" db and back
- removing first / middle / last service
- removing first / middle / last login of a service

Awesome work @deXol ... I know this took a lot.

src/MPDevice.h
deXol: To make sure not causing an issue with webauthn creds. In theory it should not, but after the mini crashes I rather made sure of it.

Issue #609, opened 2020-02-28
FAIL: build of moolticute 0.42.33-testing~xenial in ubuntu xenial RELEASE


limpkin:
Source Package: moolticute
 * Version: 0.42.33-testing~xenial
 * Architecture: amd64
 * Archive: ~mooltipass/ubuntu/moolticute-beta
 * Component: main
 * State: Failed to build
 * Duration: 2 minutes
 * Build Log: https://launchpad.net/~mooltipass/+archive/ubuntu/moolticute-beta/+build/18777428/+files/buildlog_ubuntu-xenial-amd64.moolticute_0.42.33-testing~xenial_BUILDING.txt.gz
 * Builder: https://launchpad.net/builders/lcy01-amd64-009
 * Source: not available

PR #610, opened 2020-02-28
Fix backward compatibility issue with Qt5.5.1 for Xenial build

deXol:
Qt::ISODateWithMs enum value was only added in Qt5.8
It is used for the timestamp in logging.
https://doc.qt.io/archives/qt-5.11/qdatetime.html#toString-1
> use the format Qt::ISODateWithMs, which corresponds to yyyy-MM-ddTHH:mm:ss.zzz


deXol:
Fix for #609

Issue #564, opened 2019-12-20
Updater issue in AppImage version


Dnns01:
### Expected behavior
When selecting to update moolticute to the latest version, the latest AppImage file should be downloaded

### Actual behavior
Selecting to update downloads the *.exe setup file when running AppImage version of moolticute

### Step by step guide to reproduce the problem
Download and run an AppImage, that is not the latest version, check for updates, and select to download the latest version

### Moolticute Version
42.24 testing as well as 41.5 (stable)

### Operating System
Manjaro Linux 18.1.4 (Kernel: Linux 5.4.2-1-MANJARO)

limpkin:
assigning all updating stuff to @raoulh  :)

limpkin:
current status from raoul: issue has been fixed for stable channel from 42.1, issue stilll needs to be investigated for beta channel

deXol:
**Analysis**:
Appimage version.h is created here and project is building:
https://travis-ci.org/mooltipass/moolticute/jobs/657040126#L1520
The windows one is overwriting the version.h:
https://travis-ci.org/mooltipass/moolticute/jobs/657040126#L1931

But later during packaging the windows one is used for appimage too:
https://travis-ci.org/mooltipass/moolticute/jobs/657040126#L3158
Hence the updater is searching for *.exe instead of *.appimage

limpkin:
that was a fun one....

PR #616, opened 2020-03-04
Fix AppImage update issue

deXol:
Create a copy of version.h to use it later during packeging (https://github.com/mooltipass/mc-dockers/pull/8)


deXol:
fix for #564

PR #651, opened 2020-04-22
[BLE] Implement small file storage

deXol:
### Changeset:
- Enabled Files tab
- Implemented Files Management Mode for BLE
- Added and implemented the following commands
    - MPCmd::WRITE_DATA_FILE
    - MPCmd::READ_DATA_FILE
- Disable update file button for BLE
- Update service/file name when an other file is selected

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -778,7 +778,7 @@ void MPDevice::memMgmtModeReadFlash(AsyncJobs *jobs, bool fullScan,
         }));
     }
 
-    if (getData && !isBLE()) //TODO: Data fetching is not implemented yet for BLE
+    if (getData)
     {
         //Get parent data node start address
         jobs->append(new MPCommandJob(this, MPCmd::GET_DN_START_PARENT,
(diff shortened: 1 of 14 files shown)


src/MPDevice.cpp
```diff
@@ -4245,67 +4264,85 @@ void MPDevice::getDataNode(QString service, const QString &fallback_service, con
...
+                  ));
+    }
+    else
+    {
+        jobs->append(new MPCommandJob(this, MPCmd::SET_DATA_SERVICE,
+                                      sdata,
+                                      [this, jobs, service,fallback_service](const QByteArray &data, bool &) -> bool
         {
-            if (!fallback_service.isEmpty())
+            if (pMesProt->getFirstPayloadByte(data) != 1)
```
limpkin, line 4284: magic numbers!

src/MPDevice.cpp
```diff
@@ -4385,45 +4422,74 @@ void MPDevice::setDataNode(QString service, const QByteArray &nodeData,
...
         {
-            qWarning() << "context " << service << " does not exist";
-            //Context does not exists, create it
-            createJobAddContext(service, jobs, true);
-        }
-        else
-            qDebug() << "set_data_context " << service;
-        return true;
-    }));
+            if (pMesProt->getFirstPayloadByte(data) != 1)
```
limpkin, line 4432: here as well

deXol: MPDevice contains a lot of these checks, so I added to the todo list to fix the others too.

PR #712, opened 2020-09-06
Add option to show full PIN when entering

0x0ptr:


src/MainWindow.ui
```diff
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>988</width>
-    <height>632</height>
+    <height>759</height>
```
limpkin, line 10: is that actually expanding our UI?

deXol: It is always auto generated, when you open the MainWindow.ui in Designer and set to the value based on your screen. I think we should keep the original values. (I always change them back)

deXol (approved):

src/MainWindow.ui
0x0ptr: Hi deXol!

Thanks for fixing it for me! If you prefer let me know next time and I can take care of fixing my bugs. :-)

PR #715, opened 2020-09-12
[BLE] Add tutorial enabled device setting for ble too

deXol:
Added:
kép


limpkin:
thanks! can you put the item at the bottom of the "miscellaneous" list?

Issue #711, opened 2020-09-06
Inform Mini BLE device of computer locked status


limpkin:
(feature already implemented on the mini BLE).

In short, similarly to the "lock_device 0x0010" message we send to the mini BLE whenever the computer locks itself, do the same with the "i'm locked" & "I'm unlocked" messages: https://github.com/mooltipass/minible/blob/gh-pages/protocol.md#0x0024-inform-locked
Also send the lock status at device connection.
To test if working, enable the unlock feature and see that you don't get prompted when the feature is enabled & when the computer is unlocked :)

PR #718, opened 2020-09-14
[BLE] #711 Inform device about computer is locked/unlocked

deXol:
Implement sending locked/unlocked messages when computer is locked/unlocked. #711
Also send current computer status when device is connected.
Storing computer state in MC (MainWindow) and handling lock/unlock on every OS.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -7375,7 +7375,7 @@ void MPDevice::informLocked(const MessageHandlerCb &cb)
     auto *jobs = new AsyncJobs("Inform device about computer locked", this);
 
     const auto afterFn = [](const QByteArray &, bool &) -> bool { return true; };
-    jobs->append(new MPCommandJob(this, MPCmd::LOCK_DEVICE, afterFn));
+    jobs->append(new MPCommandJob(this, MPCmd::INFORM_LOCKED, afterFn));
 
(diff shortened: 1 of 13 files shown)


limpkin:
Quick question: where are you checking it's for the BLE only? :)

deXol:
> Quick question: where are you checking it's for the BLE only? :)

**On daemon side** inform_locked and inform_unlocked messages are only processed for ble, because it implemented in `processMessageBLE` function: https://github.com/mooltipass/moolticute/pull/718/files#diff-d2c769b75b85702cadd646dcd0917fddR1171
For mini `processMessageMini` function is called, where these two messages are not added: https://github.com/mooltipass/moolticute/blob/c4a9c6dfb403c654ed1828cba4cfad5fbef0fb6a/src/WSServerCon.cpp#L508

**On GUI side** in `onDeviceConnection` isBLE is checked: https://github.com/mooltipass/moolticute/pull/718/files#diff-1c0b64226a129d35f02dc0049484cd6bR1712-R1722
However there are two place where it is not check as I see: https://github.com/mooltipass/moolticute/pull/718/files#diff-1c0b64226a129d35f02dc0049484cd6bR1634 and https://github.com/mooltipass/moolticute/pull/718/files#diff-1c0b64226a129d35f02dc0049484cd6bR1648
It will not cause any issue (because not processed on daemon side), but I will add these two extra check to avoid the overhead.

Issue #735, opened 2020-10-04
Mini BLE: implement support for password prompt bypass


limpkin:
see https://github.com/mooltipass/minible/issues/154 

The bitmask has a new element (0x20) allowing to specify password question prompt bypass. On the setting tab, I'd suggest adding a checkbox ("no password input prompt") setting that bitmask

PR #738, opened 2020-10-06
[BLE] Add new No Password Input Prompt device setting

deXol:
Implemented https://github.com/mooltipass/moolticute/issues/735 for BLE.
kép

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -337,17 +337,7 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
     ui->closeBehaviorComboBox->setVisible(false);
 #endif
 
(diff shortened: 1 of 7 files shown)


src/MainWindow.cpp
```diff
@@ -532,7 +522,6 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
 
     ui->checkBoxDebugHttp->setChecked(s.value("settings/http_dev_server").toBool());
     ui->checkBoxDebugLog->setChecked(s.value("settings/enable_dev_log").toBool());
-    ui->spinBoxDefaultPwdLength->setValue(s.value("settings/default_password_length", Common::DEFAULT_PASSWORD_LENGTH).toInt());
```
limpkin, line 535: why did this change? :)

PR #748, opened 2020-10-25
[BLE] Add TOTP Credential support

deXol:
- Added TOTP Credential dialog in MMM

kép

- Added CyoEncode lib for Base32 encode/decode

**Credential with TOTP:**
kép
View TOTP Credential button and a clock icon next to edit icon are displayed

**Credential without TOTP:**
kép

### DIFF
--- a/daemon.pro
+++ b/daemon.pro
@@ -11,7 +11,7 @@ CONFIG -= app_bundle
 
 CONFIG += c++11
 
-INCLUDEPATH += $$PWD/src $$PWD/src/MessageProtocol $$PWD/src/Mooltipass $$PWD/src/Settings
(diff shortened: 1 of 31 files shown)


limpkin:
stupid thing: can you align the left of the bottom with the left of the drop down lists?

limpkin:
also, can you send UTC time to the device now? :)

limpkin:
awesome work, as always! quick question: did you check that you could import / export databases with TOTP creds?

deXol:
> 
> 
> stupid thing: can you align the left of the bottom with the left of the drop down lists?

I can do something like that:
kép

But if I edit the credential it still looks weird:
kép

limpkin:
actually maybe center it in the middle of the input fields instead?

limpkin:
with the discard changes / confirm changes below

PR #760, opened 2020-11-07
Fix xenial build issue, add qToLittleEndian(T,void*) for QT version bellow 5.7

deXol:
https://doc.qt.io/qt-5/qtendian.html#qToLittleEndian
`Note: Since Qt 5.7, the type of the dest parameter is a void pointer.`
For xenial QT version 5.5.1 is used.


limpkin:
- does your PR assumes the host is little-endian as well?
- if so, should you #ifdef for the problematic qt version?

PR #774, opened 2020-11-26
[BLE] Add password field for uploading bundle

deXol:
image

Invalid Upload Password:
image


limpkin:
thanks! can you make the password field large enough if it needs to hold 32 hex chars? :)

PR #777, opened 2020-12-05
[BLE] Add Authentication Security Challenge

deXol:
image

Debug validation:
image

Result:
image

### DIFF
--- a/src/MPDeviceBleImpl.cpp
+++ b/src/MPDeviceBleImpl.cpp
@@ -766,6 +766,32 @@ void MPDeviceBleImpl::nihmReconditioning()
     mpDev->enqueueAndRunJob(jobs);
 }
 
+void MPDeviceBleImpl::getSecurityChallenge(const QString &key, const MessageHandlerCb &cb)
+{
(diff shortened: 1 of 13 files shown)


src/MPDeviceBleImpl.cpp
```diff
@@ -766,6 +766,32 @@ void MPDeviceBleImpl::nihmReconditioning()
...
 
+void MPDeviceBleImpl::getSecurityChallenge(const QString &key, const MessageHandlerCb &cb)
+{
+    AsyncJobs *jobs = new AsyncJobs("Send security challenge to device", this);
+
+    jobs->append(new MPCommandJob(mpDev, MPCmd::AUTH_CHALLENGE,
+                                Common::toHexArray(key),
+                                [this, cb](const QByteArray &data, bool &)
+    {
+        if (bleProt->getFirstPayloadByte(data) == MSG_FAILED )
```
limpkin, line 777: what if the first payload byte is also MSG_FAILED but the message length is longer?

deXol: it is a good point, changing it to check the size of the response

PR #783, opened 2020-12-13
[BLE] Add Hash Display setting on MC side

deXol:
image


limpkin:
can you move the setting after the "knock detecting..." setting.

deXol:
Moved:
image

PR #785, opened 2020-12-17
[BLE] Allow bundle upload when connected through USB and more than 60% battery

deXol:
image

### DIFF
--- a/src/BleDev.cpp
+++ b/src/BleDev.cpp
@@ -120,7 +120,7 @@ bool BleDev::isValidBundleFile(QFile* file) const
 
 void BleDev::on_btnFileBrowser_clicked()
 {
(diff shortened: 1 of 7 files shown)


limpkin:
could you change to "Please have your device connected through USB and fully charged" ?

src/BleDev.cpp
```diff
@@ -119,6 +120,13 @@ bool BleDev::isValidBundleFile(QFile* file) const
 
 void BleDev::on_btnFileBrowser_clicked()
 {
+    if (DeviceDetector::instance().isBt() ||
```
limpkin, line 123: change you change isBt with something more obvious, like "isBLEConnected"?

deXol:
image

PR #808, opened 2021-02-15
[BLE] #803 Only allow save settings if every settings are fetched from device

deXol:
Fix for #803


src/WSClient.cpp
```diff
@@ -512,6 +513,10 @@ void WSClient::onTextMessageReceived(const QString &message)
     {
         DeviceDetector::instance().setIsConnectedWithBluetooth(rootobj["data"].toBool());
     }
+    else if (rootobj["msg"] == "request_keyboard_layout")
+    {
+        m_settingsFetched = true;
```
limpkin, line 518: a note would be nice to mention we finish our init with keyboard layouts....

PR #817, opened 2021-03-10
[BLE] Implement FIDO Credentials tab

deXol:
FIDO Credentials page:
image
After entering FIDO management:
image
After removing a FIDO credential:
image

### DIFF
--- a/src/CredentialModel.cpp
+++ b/src/CredentialModel.cpp
@@ -167,7 +167,7 @@ TreeItem *CredentialModel::getItemByIndex(const QModelIndex &idx) const
     return static_cast<TreeItem *>(idx.internalPointer());
 }
 
-void CredentialModel::load(const QJsonArray &json)
+void CredentialModel::load(const QJsonArray &json, bool isFido)
 {
     if (json.isEmpty())
         return;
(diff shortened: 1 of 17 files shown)


src/CredentialModel.cpp
```diff
@@ -244,9 +226,34 @@ void CredentialModel::load(const QJsonArray &json)
             // Update login item address
             pLoginItem->setAddress(bAddress);
 
-            // Update login favorite
-            int iFavorite = cnode["favorite"].toInt();
-            pLoginItem->setFavorite(iFavorite);
+            if (isFido)
+            {
+                pLoginItem->setCategory(0);
```
limpkin, line 231: what's that for? :)

src/MPDevice.cpp
```diff
@@ -903,21 +906,23 @@ void MPDevice::startMemMgmtMode(bool wantData,
     startMmmJob->setTimeout(15000); //We need a big timeout here in case user enter a wrong pin code
     jobs->append(startMmmJob);
 
+    //Fetch credential when not data and not fido wanted
+    bool wantCredentials = !wantData && !wantFido;
```
limpkin, line 910: very nice

limpkin (commented):
very nice code! see my comments. Also:
- on an empty DB I get both "Discard deletions" & "Save & Quit" buttons... should we only have "quit" like for standard MMM?

src/CredentialModel.cpp
deXol: Without that category value is not initialized and it had have a value >0 so the category icon was displayed.

PR #827, opened 2021-03-30
[BLE] Implement reset to default setting values option, fix #823

### LINKED ISSUE #823: Feature request for the BLE: reset default settings
Idea: add a button to reset the settings to the default values, located here: https://github.com/mooltipass/minible/blob/master/source_code/main_mcu/src/FILESYSTEM/custom_fs.c#L43

deXol:
Added a `Reset default values` button, which is reseting settings values.
resetToDefaultSettings

### DIFF
--- a/src/Settings/DeviceSettingsBLE.cpp
+++ b/src/Settings/DeviceSettingsBLE.cpp
@@ -1,9 +1,44 @@
 #include "DeviceSettingsBLE.h"
 
+DeviceSettingsBLE::DefaultValues DeviceSettingsBLE::m_bleDefaultValues =
+    {
+        {MPParams::RANDOM_INIT_PIN_PARAM, 0},
+        {MPParams::USER_INTER_TIMEOUT_PARAM, 15},
+        {MPParams::FLASH_SCREEN_PARAM, 1},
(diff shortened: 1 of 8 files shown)


src/Settings/DeviceSettingsBLE.cpp
```diff
@@ -1,11 +1,46 @@
...
     : DeviceSettings(parent)
 {
     fillParameterMapping();
 }
 
+void DeviceSettingsBLE::resetDefaultSettings()
+{
+    for (DefaultValues::iterator it = m_bleDefaultValues.begin(); it != m_bleDefaultValues.end(); ++it)
+    {
+        updateParam(it.key(), it.value());
```
limpkin, line 40: is this triggering 20 or so messages? this is not really ideal as far as memory writes go...

deXol: No, it is just updating the values on daemon side. Most device settings are stored with SET_DEVICE_SETTINGS (0x000d) message which is sent only once for the device during reset.

limpkin: awesome!

PR #830, opened 2021-04-05
[BLE] Add enforce keyboard layout option, fix #807

### LINKED ISSUE #807: Feature request: force keyboard layout on a given moolticute install
Using newer 0x0037 HID command every time the user is logged in

deXol:
Added "Enforce at connection" checkbox for keyboard layouts:
image
If they are enabled then after user login when layouts are fetched sending the new 0x0037 message to enforce the layout.


deXol:
todo: Tooltip about details and information about enforce at connection is only stored on MC side

PR #834, opened 2021-05-27
[BLE] Implement Notes tab

deXol:
Added a new tab for Notes, enabled for bundle version >= 1.
image

Demo:
notesDemo

Missing functionality:
Export/Import notes


src/MPDevice.cpp
```diff
@@ -1151,6 +1156,54 @@ void MPDevice::loadSingleNodeAndScan(AsyncJobs *jobs, const QByteArray &address,
...
+            /* Delete empty parent
+             * We do not need CRED_ADDR_IDX, but cannot call without it
+             */
+
+            removeEmptyParentFromDB(parentPt, true, Common::CRED_ADDR_IDX, addrType);
+        }
+    }
+}
+
+bool MPDevice::checkDataNodeChanged(AsyncJobs *jobs, Common::DataAddressType addrType)
```
limpkin, line 1186: as this function actually performs an action, could you rename it? maybe prepend "AndGeneratePacket"?

src/MPDevice.cpp
```diff
@@ -1750,8 +1820,21 @@ void MPDevice::detagPointedNodes(Common::AddressType addrType /*= Common::CRED_A
...
+    for (auto &i: notesLoginChildNodes)
+    {
+        i->removePointedToCheck();
+    }
+}
+
 /* Follow the chain to tag pointed nodes (useful when doing integrity check when we are getting everything we can) */
-bool MPDevice::tagPointedNodes(bool tagCredentials, bool tagData, bool repairAllowed, Common::AddressType addrType /*= Common::CRED_ADDR_IDX*/)
+bool MPDevice::tagPointedNodes(bool tagCredentials, bool tagData, bool repairAllowed, Common::AddressType addrType /*= Common::CRED_ADDR_IDX*/,
+                               Common::DataAddressType dataAddrType /*= Common::DATA_ADDR_IDX*/)
```
limpkin, line 1837: that may be a stupid question: have you checked that we never call tagPointedNodes with true, true? I mean, is there a use case where we need to tag all data types & credential types at once, as this code wouldn't be able to do it?

src/MPDeviceBleImpl.cpp
```diff
@@ -516,22 +628,33 @@ bool MPDeviceBleImpl::processReceivedData(const QByteArray &data, QByteArray &da
...
+    qDebug() << "Received starting node size: " << res.size();
+    if (AppDaemon::isDebugDev())
+    {
+       qDebug() << "Starting addresses: " << res;
+    }
     return res;
  }
 
- QByteArray MPDeviceBleImpl::getDataStartNode(const QByteArray &data) const
+ QVector<QByteArray> MPDeviceBleImpl::getDataStartNode(const QByteArray &data) const
```
limpkin, line 639: should we rename it getDataStartNodes?

src/MPDevice.cpp
```diff
@@ -2156,13 +2247,22 @@ bool MPDevice::tagPointedNodes(bool tagCredentials, bool tagData, bool repairAll
     return return_bool;
 }
 
-bool MPDevice::deleteDataParentChilds(MPNode *parentNodePt)
+bool MPDevice::deleteDataParentChilds(MPNode *parentNodePt, Common::DataAddressType addrType /*= Common::DATA_ADDR_IDX*/)
 {
+    const bool isData = Common::DATA_ADDR_IDX == addrType;
+    NodeList& childNodes = isData? dataChildNodes : notesLoginChildNodes;
```
limpkin, line 2253: can you rename that variable? that may get confusing quickly...

src/MPDevice.cpp
```diff
@@ -2600,13 +2702,20 @@ bool MPDevice::removeEmptyParentFromDB(MPNode* parentNodePt, bool isDataParent,
     }
 
     /* Detag them */
-    detagPointedNodes(addrType);
+    if (isDataParent && !isData)
+    {
+        detagPointedNotesNodes();
```
limpkin, line 2707: same question, why not add the feature to detagPointedNodes instead?

src/MPDevice.cpp
limpkin: actually it seems it happens in checkloadednodes in:

if (checkCredentials || checkData)
          {
              return_bool &amp;= tagPointedNodes(checkCredentials, checkData, repairAllowed);
          }

limpkin: actually you kind of do it after with return_bool &= tagPointedNodes(true, checkData, repairAllowed, Common::WEBAUTHN_ADDR_IDX);

this is getting a tad complex IMHO... should we change tagPointedNodes to not use Common::AddressType addrType & Common::DataAddressType dataAddrTyp and use two bools instead?

src/MPDevice.cpp
```diff
@@ -2989,33 +3101,12 @@ bool MPDevice::checkLoadedNodes(bool checkCredentials, bool checkData, bool repa
...
-                if (repairAllowed)
-                {
-                    addOrphanParentToDB(i, true, repairAllowed);
-                }
-                nbOrphanDataParents++;
-            }
-        }
-        QListIterator<MPNode*> i(dataChildNodes);
-        while (i.hasNext())
+        checkLoadedDataNodes(nbOrphanDataParents, nbOrphanDataChildren, repairAllowed, Common::DATA_ADDR_IDX);
```
limpkin, line 3104: the tag pointednode check is doing at f (checkCredentials && !tagPointedNodes(checkCredentials, checkData, repairAllowed)) ?

can you explain why you specified false in the following call?

tagPointedNodes(false, true, repairAllowed, Common::CRED_ADDR_IDX, Common::NOTE_ADDR_IDX);

src/MPDevice.cpp
```diff
@@ -4669,33 +4850,18 @@ void  MPDevice::deleteDataNodesAndLeave(QStringList services,
...
-
-                /* Delete empty parent */
-                removeEmptyParentFromDB(parentPt, true);
-            }
+            deleteDataNodes(notes, cb, Common::NOTE_ADDR_IDX);
         }
 
         /* Check our DB */
-        if(!checkLoadedNodes(false, true, false))
+        if(!checkLoadedNodes(false, true, false, false, true))
```
limpkin, line 4864: nice!

limpkin (changes requested):

src/MPDevice.cpp
```diff
@@ -1971,18 +2062,18 @@ bool MPDevice::tagPointedNodes(bool tagCredentials, bool tagData, bool repairAll
         }
     }
 
-    if (tagData && isCred)
```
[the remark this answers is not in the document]
deXol: isCred is `const bool isCred = Common::CRED_ADDR_IDX == addrType;`, so only not checked data nodes when it was called for Webauthn nodes which should not happen.

src/MPDevice.cpp
deXol: > (checkCredentials && !tagPointedNodes(checkCredentials, checkData, repairAllowed)) ?
tagPointedNodes in this call is tagging pointed nodes for data (files) nodes, because the last parameter is Common::DATA_ADDR_IDX by default.

> 
> can you explain why you specified false in the following call?
> tagPointedNodes(false, true, repairAllowed, Common::CRED_ADDR_IDX, Common::NOTE_ADDR_IDX);
In this case we are tagging pointed nodes for Notes, the first parameter is tagCredentials, which we should not tag in this case.

src/MPDevice.cpp
```diff
@@ -3396,6 +3469,99 @@ bool MPDevice::generateSavePackets(AsyncJobs *jobs, bool tackleCreds, bool tackl
...
+    {
+        /* See if we can find the same node in the clone list */
+        temp_node_pointer = findNodeWithAddressInList(dataNodesListClone, nodelist_iterator->getAddress(), 0);
+
+        if (!temp_node_pointer)
+        {
+            qDebug() << "Generating save packet for new data service" << nodelist_iterator->getService();
+            addWriteNodePacketToJob(jobs, nodelist_iterator->getAddress(), nodelist_iterator->getNodeData(), writeCb);
+            diagSavePacketsGenerated = true;
+            progressTotal += 3;
```
limpkin, line 3491: that += 3 may only be for the mini...

PR #868, opened 2021-06-22
Add AppStream metainfo file

principis:
This adds support for software centers. [More information](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps). 

If moolticute is available in a distribution's repository, the metainfo makes sure users can find it within their software centers (gnome-software, discover, ...). This is useful if they are not familiar with command line tools.

A lot more information could be added. For example screenshots (they need to be available on a url). [Gajim](https://dev.gajim.org/gajim/gajim/-/blob/master/data/org.gajim.Gajim.appdata.xml.in) is a good example which uses a lot of the features.

gnome-software example

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,12 @@ test:
 	cd build; $(MAKE) check
 
 install:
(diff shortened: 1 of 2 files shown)


limpkin:
thanks a lot @principis  :)

PR #882, opened 2021-07-03
Added additional fedora dnf intall command.

reddraconi:
Fedora 33 and > uses `qt5-qtbase` rather than `qt5` for the qt5 base package.
I wasn't sure the best way to change the README.md file while still making it copy-and-pastable, so I just added the updated
string in a new code fence.


limpkin:
thanks a lot :)

PR #892, opened 2021-07-26
[BLE] Implement Integrity Check

deXol:
Implemented integrity check for BLE and updated test cases for it.


src/MPDevice.cpp
```diff
@@ -965,6 +965,11 @@ void MPDevice::startMemMgmtMode(bool wantData, bool wantFido,
 
 QByteArray MPDevice::getMemoryFirstNodeAddress(void)
 {
+    if (isBLE())
+    {
+        /* BLE: Address format is 2 bytes little endian. last 1 bit is node number (for pages that can accomodate 2 nodes) and first 15 bits are page address */
+        return QByteArray::fromHex("0200");
```
limpkin, line 971: I don't think this is correct.
Address format is page << 1
first address available (from the firmware code):
image
so for the 1M / 2M / 32M version it is 0100

limpkin:
beautiful! 
Can you just review my comment and modify if I'm right?

PR #909, opened 2021-08-28
Display warning notification when battery is low, fix #801

### LINKED ISSUE #801: Feature request: display warning when battery is too low

deXol:
image
Display notification when battery is below 30% and not charging.


limpkin:
i think the user will want to discard it... can you implement the way to do so? :)

deXol:
It is a system notification :) It will disappear after few seconds of if the user hovers on it can click on X.

Issue #943, opened 2021-12-04
"Tutorial is only available for ...." appears several times + all tabs are disabled


cwaazywabbit:
I am unable to use the app as usual and receive a warning several times.

### Expected behavior
1. Not seeing that warning (at least not several times)
2. Being able to use the app as usual

### Actual behavior
1. I receive the warning several times when opening the app

```
Tutorial is only available for BLE device, please attach one or Exit tutorial.
```

2. I am able to use the app. All tabs are greyed out (disabled).

### Moolticute Version
Tested with 0.53.1 and the version before. I don't know which version would work, please advice.

### Operating System
Windows 10 Build 17763

### Mooltipass Device
- The Mooltipass Mini

Please let me know which version would/should work, so I can at least use that until the problem is fixed.

PR #944, opened 2021-12-04
Exit MC tutorial when Mini is connected

deXol:


limpkin:
fix #943

cwaazywabbit:
Thanks for the quick fix. Are binaries being generated and will be available in the assets later?

PR #969, opened 2022-01-16
Change rand calls to QRandomGenerator

deXol:
Changing to [QRandomGenerator](https://doc.qt.io/qt-5/qrandomgenerator.html), but we still need to keep the `rand` calls for below Qt 5.10 versions.


limpkin:
tagging @raoulh

PR #972, opened 2022-01-17
Fix Gnome on Wayland crash

### LINKED ISSUE #872: cant start moolticute on Fedora34 with Wayland
Build from source:

0.43.3 is starting.
0.46.1 is throwing error

```
WARNING: (2021-06-24T12:06:03.793) :0 - The Wayland connection experienced a fatal error: Protocol error
```

its is starting with sudo - but thats not the goal.
file have same permissions. 
So the master-question - what has changed since than ;)

principis:
On some Gnome on Wayland platforms, moolticute wouldn't launch with a
protocol error: "Invalid min/max size". This seems to be caused by an
invalid maximum width value. This patch resets this value.

Fixes #872

### DIFF
--- a/src/AppGui.cpp
+++ b/src/AppGui.cpp
@@ -792,6 +792,12 @@ void AppGui::createMainWindow()
     Q_ASSERT(dbMasterController);
 
     win = new MainWindow(wsClient, dbMasterController);
+
+    // Force resize and reset maximumWidth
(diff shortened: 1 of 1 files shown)


limpkin:
waiting for input from @deXol before merging this one

principis:
I did some more testing and this breaks remembering the last window size on launch.

Issue #961, opened 2022-01-04
Moolticute window size too large


limpkin:
From one of our customers:

I'm currently noticing that the bottom part (specific the part of the window where the acknowledge buttons reside) from the moolticute app (in Linux and in Windows) falls below the visible window/screen size. I am not able to resize the window to make the lower part visible. This in spite of the screen resolution being 1366*768 or full hd. In that specific part of the screen the confirm or cancel button and the categories (in credentials) reside and thus cannot select them.
Distribution: Debian Linux Bullseye

![image](https://user-images.githubusercontent.com/6170291/148108588-27869031-c63a-48e5-bfd1-a733963696cd.png)
![image](https://user-images.githubusercontent.com/6170291/148108674-d0f46df2-59b7-4136-b986-71719f197f56.png)

Kaiserdragon2:
I have the same Issue on Windows 10 21H1 with a screen resolution of 3840x2160 and scaling set to 250%.

Kaiserdragon2:
![2022-01-09](https://user-images.githubusercontent.com/8929967/149989117-04b22f42-1998-4790-9be1-b6d1dd563d28.png)

Kaiserdragon2:
While trying to record the screen to get a video of the window view runing out of the view while trying to resize, now it works normall again. It also worked normal some time before.

limpkin:
currently building v0.53.13-testing with this fix!
For windows / macOS it will be available at betas.themooltipass.com/
On ubuntu https://launchpad.net/~mooltipass/+archive/ubuntu/moolticute-beta

ghalfacree:
I can confirm the window is smaller in v0.53.13-testing: I've measured its minimum height at 631 pixels including window decoration, in Ubuntu 20.04.

limpkin:
awesome!

Kaiserdragon2:
![grafik](https://user-images.githubusercontent.com/8929967/152506619-eb737fff-457a-4258-92ca-7e7be9c70ba1.png)

It's better now on Windows but a small bit is still outside the view and I am not able to resize the window to a smaller size.

Kaiserdragon2:
![grafik](https://user-images.githubusercontent.com/8929967/152509995-8e4b1ca6-262c-40d7-96f1-460979f7b6b6.png)
If the hidden tabs are active the window is now to wide and i am not able to resize the width at all

limpkin:
That's cutting it close... but I think it's still quite acceptable no? :)

[...]

PR #981, opened 2022-02-02
Decrease minimum height of Moolticute

deXol:
Put Moolticute settings tab content to a scroll area:
kép
Removed section texts for Credential tabs:
kép

With these changes MC's min height with banner on the bottom is 700px.


limpkin:
fix #961

PR #987, opened 2022-02-12
Detect Focus Assist for Windows

deXol:
Implemented solution from https://stackoverflow.com/questions/53407374/is-there-a-way-to-detect-changes-in-focus-assist-formerly-quiet-hours-in-windo/53411482#53411482


limpkin:
you rock :)

PR #995, opened 2022-03-03
Fix delete and rename account to the same name in MMM, fix #706

### LINKED ISSUE #706: Unable to delete account and rename account at same time
### Expected behavior
In Moolticute management mode, I've got two accounts, "account" and "account - new". I would like to remove "account" and rename "account - new" to be just "account". I used this paradigm when migrating a password for a really sensitive database so that if something happened, everything would revert, and I'd still have old password available.

### Actual behavior
There is an error from Moolticute in the logs. Turns out, Moolticute doesn't really like deleting one index and then renaming another to be the same. See redacted output below:

DEBUG: (2020-09-01T19:21:52.888) :0 - Removing child  "account"  with parent  "[REDACTED]"  from DB
INFO: (2020-09-01T19:21:52.888) :0 - Checking database...
WARNING: (2020-09-01T19:21:52.891) :0 - Orphan child found: "account" at address: "[REDACTED ADDRESS]"
DEBUG: (2020-09-01T19:21:52.891) :0 - Number of parent orphans: 0
DEBUG: (2020-09-01T19:21:52.891) :0 - Number of children orphans: 1
DEBUG: (2020-09-01T19:21:52.891) :0 - Number of data parent orphans: 0
DEBUG: (2020-09-01T19:21:52.891) :0 - Number of data children orphans: 0
INFO: (2020-09-01T19:21:52.891) :0 - Errors were found in the database
CRITICAL: (2020-09-01T19:21:52.891) :0 - Error in our local DB (algo PB?)

I'm not sure if there is anything that can really be done to accommodate this set of steps aside from document around it or maybe find a better way to detect that its happening and give a more useful error.

### Step by step guide to reproduce the problem
With any given parent, create a child of "account" and "account - new". Go in to credential management mode and in one session, remove "account" and rename "account - new" to be "account" Click save all and behold the error.

### Moolticute Version
v0.44.0

### Operating System

- Windows 7/10

### Mooltipass Device
- The Mooltipass Mini

deXol:
Repro steps in #706 
During renaming credential addChildToDB call for the newNode fails, because the deleted node has not been removed yet.
Later deleted credentials are processed and a check for the potentially changed nodes handled here.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -7514,6 +7514,8 @@ void MPDevice::setMMCredentials(const QJsonArray &creds, bool noDelete,
     newAddressesReceivedCounter = 0;
     bool packet_send_needed = false;
     bool onlyChangePwd = isBLE() && isCsv && !bleImpl->get_advancedMenu();
+    //New node ptr and parent address for potentially changed nodes
+    QHash<MPNode*, QByteArray> potentialChangedNodes;
     AsyncJobs *jobs = new AsyncJobs("Merging credentials changes", this);
 
     /// TODO: sanitize inputs (or not, as it is done at the mpnode.cpp level)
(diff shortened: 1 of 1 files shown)


limpkin:
nice work!

PR #1015, opened 2022-03-21
Refactor common issue template to separate issue templates

gmarcusm:
No need to have a combined issue templates, define one template for bug reports and another for feature requests.


limpkin:
thanks a lot! I had implemented it in the mini ble repo but never here :)

PR #1016, opened 2022-03-24
Add duplicate service check for import and leaving MMM, fix #1011

### LINKED ISSUE #1011: To investigate: a possible situation where moolticute creates a duplicated service name
Following https://github.com/mooltipass/minible/issues/320 I wonder if there's not a situation where moolticute is going to create a duplicate service. 
I could imagine a case where one backup file contains 2 logins for a service, and the user DB having only one.... to be tested.
Perhaps we should check in moolticute (during the database check operations) for duplicated services? 
To check if a mini BLE has a duplicated service: go to the login menu, press and scroll up. If it reboots, then the DB has an issue.

deXol:


limpkin:
congrats!

Issue #953, opened 2021-12-18
[improvement] Moolticute should display notes using a monospaced font.


alanrick:
Use one of the two templates below and delete the other. All issues not following this template will be directly closed.

8<------------------------ BUG REPORT -----------------------------------------
### Expected behavior
Moolticute improvement.

https://www.reddit.com/r/mooltipass/comments/681bd7/banking_passwords/

> Several high street banks use a zero-knowledge authentication to avoid keyloggers. 
> Basically you don't type in your password but instead you are challenged to enter specific digits from your password. E.g. 4th and 8th and 11th digits...

I can think of other use-cases where vertically spaced columns are useful in the Notes tab. E.g. list of credit card pins.

### Actual behavior
The current behavior of showing the text in the Note tab in a proportional font prevents this technique.

> Userid  peterpiper44
> 12345 67890 12345 67890 12345 67890
> mm345 12jdk 34561 lllll jjjjj mabcd

as opposed to the preferred behavior:

```
Userid  peterpiper44
12345 67890 12345 67890 12345 67890
mm345 12jdk 34561 lllll jjjjj mabcd
```
which is far easier to interpret.

### Moolticute Version
0.53.7

### Operating System
Mention if you are using either:
- Windows 10

### Mooltipass Device

- The Mooltipass Mini BLE

8<------------------------ END BUG REPORT -------------------------------------

PR #1024, opened 2022-04-07
Issue #953 Monospaced notes

alanrick:
## Fixes mooltipass/moolticute#953
Using the _courier_ font in the Notes editor.

The notesmanagement.ui file was changed in the QT designer instead of a text editor, which generated a _height_ change and increment of the _stackedWidget_ _number_.

Tested in WIndows 10 only.


alanrick:
My first ever pull request, **so worth reviewing.**  But couldn't figure how to flag this.
The only issue I foresee is that the courier font might not be supported in the Linux environments. I used the designer in QT in the hopes that this handles compatibility.

PR #1028, opened 2022-04-20
[BLE] Implement Get TOTP code command

deXol:
Implemented GET_TOTP_CODE (0x0041) command on daemon side.
Message "get_totp_code" is processed with "service", "login" data and totp code is sent back with "totp_code" data.

### DIFF
--- a/src/MPDeviceBleImpl.cpp
+++ b/src/MPDeviceBleImpl.cpp
@@ -467,7 +467,7 @@ void MPDeviceBleImpl::getCredential(const QString& service, const QString& login
         jobs = new AsyncJobs(getCred, reqid, this);
     }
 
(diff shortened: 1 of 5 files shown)


src/MPDeviceBleImpl.cpp
```diff
@@ -467,7 +467,7 @@ void MPDeviceBleImpl::getCredential(const QString& service, const QString& login
         jobs = new AsyncJobs(getCred, reqid, this);
     }
 
-    jobs->append(new MPCommandJob(mpDev, MPCmd::GET_CREDENTIAL, createGetCredMessage(service, login),
+    jobs->append(new MPCommandJob(mpDev, MPCmd::GET_TOTP_CODE, createGetCredMessage(service, login),
```
limpkin, line 470: i'm guessing this is not wanted? :D

deXol: oh, nope  (this remained from testing :D)

PR #1031, opened 2022-04-26
Minimize MC with cmd + Q on Mac

deXol:
Remove option to customize cmd + Q for terminate application. Now it will only minimize Moolticute.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -388,46 +388,13 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
 
     // Close behavior
 #ifdef Q_OS_MAC
-    ui->closeBehaviorComboBox->addItem(
-        tr("Close Application"), static_cast<int>(CloseBehavior::CloseApp));
-    ui->closeBehaviorComboBox->addItem(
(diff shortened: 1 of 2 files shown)


rastagraffix:
> Remove option to customize cmd + Q for terminate application. Now it will onl
> y minimize Moolticute.

What on earth are you doing?  CMD-Q is the bog standard One True Way
to QUIT an app.  Don't mess with that.  CMD-M is the one you use to
minimize something.

limpkin:
@rastagraffix we recently had the case were a customer was unwillingly closing moolticute using cmd-q and wondering why he couldn't use native integration anymore.
We then wondered what would be the reason for a user to close moolticute other than by using the dedicated menu in the tray (as for windows)

rastagraffix:
Mathieu writes:
>
> @rastagraffix we recently had the case were a customer was unwillingly closin
> g moolticute using cmd-q and wondering why he couldn't use native integration
>  anymore.

That's just operator error, and the fix is judicious application of
the clue bat.  CMD-Q is QUIT, and has been since the first version of
MacOS shipped.  Don't mess with that.  Please revert this change.

--lyndon

PR #1033, opened 2022-05-01
[BLE] Add Current Category setting

deXol:
kép
- If a category is selected setting it as current category, when BLE device is connected and unlocked.
- Using category names from connected BLE device's categories.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -226,6 +226,7 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
                 auto currentCategoryIndex = ui->comboBoxBleCurrentCategory->currentIndex();
                 if (currentCategoryIndex != 0)
                 {
+                    // When enforce current category is set sending category number to device
                     wsClient->sendCurrentCategory(currentCategoryIndex);
                 }
             }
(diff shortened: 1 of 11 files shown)


src/MainWindow.cpp
```diff
@@ -223,6 +223,11 @@ MainWindow::MainWindow(WSClient *client, DbMasterController *mc, QWidget *parent
                     m_notesFetched = true;
                     wsClient->sendFetchNotes();
                 }
+                auto currentCategoryIndex = ui->comboBoxBleCurrentCategory->currentIndex();
+                if (currentCategoryIndex != 0)
```
limpkin, line 227: explanation here ;)

src/MainWindow.cpp
```diff
@@ -2324,3 +2348,28 @@ void MainWindow::sendRequestNotes(int bundle)
...
+    }
+    QSettings s;
+    ui->comboBoxBleCurrentCategory->blockSignals(true);
+    ui->comboBoxBleCurrentCategory->clear();
+    ui->comboBoxBleCurrentCategory->addItem(tr("Disable"), 0);
+    ui->comboBoxBleCurrentCategory->addItem(cat1, 1);
+    ui->comboBoxBleCurrentCategory->addItem(cat2, 2);
+    ui->comboBoxBleCurrentCategory->addItem(cat3, 3);
+    ui->comboBoxBleCurrentCategory->addItem(cat4, 4);
+    ui->comboBoxBleCurrentCategory->setCurrentIndex(s.value("settings/current_category", 0).toInt());
```
limpkin, line 2373: rename the setting to enforced_category maybe?

PR #1048, opened 2022-06-11
Add SIL OFL 1.1 license

principis:
`src/QtAwesome/QtAwesome/fonts/fontawesome-4.6.1.ttf` is licensed under SIL OFL 1.1.
This license is missing from the repository, but should be included.

I was notified about the license missing from the Fedora package. Since this should be included in the tarball anyway, I've added it.


limpkin:
nice catch! thanks a lot!

PR #1062, opened 2022-07-31
[BLE] Mirroring implementation

deXol:
Implemented password mirroring support in MMM.
Demo:
mirrorDemo


src/CredentialModel.cpp
```diff
@@ -363,6 +373,27 @@ QSet<qint8> CredentialModel::getTakenFavorites() const
     return favorites;
 }
 
+QString CredentialModel::getCredentialNameForAddress(QByteArray addr) const
```
limpkin, line 376: comments welcome in the future ;)

limpkin (approved):
wow that must have been quite something to work on... nice work! congrats :)

PR #1063, opened 2022-08-07
[BLE] Fix get/setLoginParent/ChildNodeData

deXol:
For parent node "last child node used addr" and for child node "pointed to child address" was included.

### DIFF
--- a/src/Mooltipass/MPNode.cpp
+++ b/src/Mooltipass/MPNode.cpp
@@ -334,36 +334,13 @@ QByteArray MPNode::getNodeFlags() const
     return data.mid(NODE_FLAG_ADDR_START, ADDRESS_LENGTH);
 }
 
-QByteArray MPNode::getLoginNodeData() const
-{
-    // return core data, excluding linked lists and flags
-    if (!isValid()) return QByteArray();
-    return data.mid(DATA_ADDR_START);
-}
-
-void MPNode::setLoginNodeData(const QByteArray &flags, const QByteArray &d)
-{
-    // overwrite core data, excluding linked lists
-    if (isValid())
-    {
-        data.replace(DATA_ADDR_START, pMesProt->getParentNodeSize()-DATA_ADDR_START, d);
-        data.replace(0, ADDRESS_LENGTH, flags);
-    }
-}
-
-QByteArray MPNode::getLoginChildNodeData() const
-{
-    // return core data, excluding linked lists and flags
-    if (!isValid()) return QByteArray();
-    return data.mid(LOGIN_CHILD_NODE_DATA_ADDR_START);
-}
-
 void MPNode::setLoginChildNodeData(const QByteArray &flags, const QByteArray &d)
 {
     // overwrite core data, excluding linked lists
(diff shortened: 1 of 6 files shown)


src/Mooltipass/MPNode.cpp
```diff
@@ -334,36 +334,13 @@ QByteArray MPNode::getNodeFlags() const
...
-    return data.mid(LOGIN_CHILD_NODE_DATA_ADDR_START);
-}
-
 void MPNode::setLoginChildNodeData(const QByteArray &flags, const QByteArray &d)
 {
     // overwrite core data, excluding linked lists
     if (isValid())
     {
-        data.replace(LOGIN_CHILD_NODE_DATA_ADDR_START, pMesProt->getChildNodeSize()-LOGIN_CHILD_NODE_DATA_ADDR_START, d);
+        const int childNodeDataAddrStart = getLoginChildNodeDataAddrStart();
```
limpkin, line 342: That var name is confusing... maybe append "Offset" to it?

src/Mooltipass/MPNodeBLE.cpp
```diff
@@ -35,7 +35,33 @@ bool MPNodeBLE::isValid() const
...
     return (data.size() == PARENT_NODE_LENGTH && (NodeParent == type || NodeParentData == type)) ||
-           (data.size() == CHILD_NODE_LENGTH && (NodeChild == type || NodeChildData == type));
+            (data.size() == CHILD_NODE_LENGTH && (NodeChild == type || NodeChildData == type));
+}
+
+QByteArray MPNodeBLE::getLoginNodeData() const
+{
+    // return core data, excluding linked lists, flags and last child node used address
+    if (!isValid()) return QByteArray();
+    return data.mid(DATA_ADDR_START, LAST_CHILD_NODE_USED_ADDR_START - DATA_ADDR_START);
```
limpkin, line 45: I don't get that one at all, when it used to be 
`return data.mid(DATA_ADDR_START);`

src/Mooltipass/MPNodeBLE.cpp
```diff
@@ -35,7 +35,33 @@ bool MPNodeBLE::isValid() const
...
+}
+
+void MPNodeBLE::setLoginNodeData(const QByteArray &flags, const QByteArray &d)
+{
+    // overwrite core data, excluding linked lists and last child node used address
+    if (isValid())
+    {
+        const auto parentNodeSize = pMesProt->getParentNodeSize();
+        const auto bytesAfterLastUsed = parentNodeSize - LAST_CHILD_NODE_USED_ADDR_START;
+        data.replace(DATA_ADDR_START, parentNodeSize - DATA_ADDR_START - bytesAfterLastUsed, d);
```
limpkin, line 55: before the second arg would be `pMesProt->getParentNodeSize()-DATA_ADDR_START` but now it is LAST_CHILD_NODE_USED_ADDR_START - DATA_ADDR_START  ?

limpkin:
nice work!
fixes double-created service in DB.

PR #1065, opened 2022-08-27
[BLE] Add last used child node and pointed to address check for import

deXol:
Check the addresses from the imported DB and find the correct service/login for them in the device's DB.

### DIFF
--- a/src/MPDevice.cpp
+++ b/src/MPDevice.cpp
@@ -6176,6 +6176,11 @@ void MPDevice::startImportFileMerging(const MPDeviceProgressCb &cbProgress, Mess
                 /* We got all the addresses, change virtual addrs for real addrs and finish merging */
                 changeVirtualAddressesToFreeAddresses();
 
+                if (isBLE())
+                {
+                    checkImportedLoginLastChildNodeUsedAddr();
+                }
+
                 /* Finish import file merging */
                 QString stringError;
                 if(finishImportFileMerging(stringError, noDelete))
@@ -6255,6 +6260,10 @@ void MPDevice::startImportFileMerging(const MPDeviceProgressCb &cbProgress, Mess
         }
         else
         {
+            if (isBLE())
+            {
+                checkImportedLoginLastChildNodeUsedAddr();
+            }
             /* Finish import file merging */
             QString stringError;
             if(finishImportFileMerging(stringError, noDelete))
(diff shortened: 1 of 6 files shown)


src/MPDevice.cpp
```diff
@@ -6766,6 +6779,119 @@ bool MPDevice::checkImportedDataNodes(const MessageHandlerCb &cb, Common::DataAd
...
+                // Credential parent node is found in the imported list
+                if (serviceName == importedLoginNodes[i]->getService())
+                {
+                    auto* importedLoginNode = static_cast<MPNodeBLE*>(importedLoginNodes[i]);
+                    // If the imported last child node used address is set, find the same login in local list
+                    if (Common::NOT_SET_ADDR != importedLoginNode->getLastChildNodeUsedAddr())
+                    {
+                        // First find the login name of the last child node used from imported list
+                        auto* importedLastUsedChildNode = findNodeWithAddressWithGivenParentInList(importedLoginChildNodes, importedLoginNode, importedLoginNode->getLastChildNodeUsedAddr(), 0);
+                        if (importedLastUsedChildNode)
```
limpkin, line 6809: should we make an else() with a qwarning ?

src/MPDevice.cpp
```diff
@@ -6766,6 +6779,119 @@ bool MPDevice::checkImportedDataNodes(const MessageHandlerCb &cb, Common::DataAd
...
+                        {
+                            auto lastUsedChildName = importedLastUsedChildNode->getLogin();
+                            qDebug() << "Last used child name: " << lastUsedChildName;
+                            // Find the node with the same login name for the given service
+                            auto* lastUsedChild = findNodeWithLoginWithGivenParentInList(loginChildNodes, loginNode, lastUsedChildName);
+                            if (lastUsedChild && lastUsedChild->getAddress() != bleLoginNode->getLastChildNodeUsedAddr())
```
limpkin, line 6815: nice! but maybe an else for the lastUsedChild condition, as it'd mean we messed up?

src/MPDevice.cpp
```diff
@@ -6766,6 +6779,119 @@ bool MPDevice::checkImportedDataNodes(const MessageHandlerCb &cb, Common::DataAd
...
+            while (childAddr != Common::NOT_SET_ADDR)
+            {
+                auto* childNode = findNodeWithAddressInList(loginChildNodes, childAddr);
+                // If login is tagged for merge check the pointed to address for the node
+                if (childNode && childNode->getMergeTagged())
+                {
+                    auto* bleChildNode = static_cast<MPNodeBLE*>(childNode);
+                    // Find the child node for the same login name in imported list
+                    auto* importedChildNode = findNodeWithLoginWithGivenParentInList(importedLoginChildNodes, importedLoginNode, childNode->getLogin());
+                    if (importedChildNode)
```
limpkin, line 6861: no else() needed here as we may do an additive merge.. though we could also check that merge condition

limpkin (approved):
very nice.

Issue #1052, opened 2022-06-14
Unlock Windows 10 computer is not fully working


fc-projects:
### Expected behavior
Enter my pin on Mooltipass, the computer is unlocked with the password stored in the _unlock_ entry.

### Actual behavior
It partially works, only if no key of the keyboard has been hit before entering the Mooltipass pin.
If any key has been hit, Windows 10 is entering the password field, and then when the Mooltipass tries to gain access to the password field by pressing ENTER it validates an empty password which fails to log in.

### Step by step guide to reproduce the problem
Configure the unlock feature with "Enter + password" or "Enter + password / Win + L".
If not done, create in the Mooltipass an _unlock_ entry with your password for opening your Windows session.
Hit any key on the keyboard, for example CTRL then try to unlock your session with the unlock feature.
It will not work (unless your password is actually empty).

Note: as far as I know there is no way on Windows 10 to go back to the locked state (presenting the picture), otherwise this would have been a valid and useful workaround.

Proposed solution: add an option to use CTRL key to gain access to the password field which works also when the field has already the focus instead of ENTER which obviously does not.

### Moolticute Version
v0.53.7

### Operating System
Windows 10

### Mooltipass Device
Mooltipass Mini BLE

fc-projects:
I digged around into the code of Moolticute and Mooltipass mini BLE and did understand how the Lock/Unlock feature actually works. I was not aware that it was the Mooltipass which was typing also the additionnal keys before and after.
Then implementing my request would need a modification in both Moolticute and embedded firmware of devices.

By seeing the code I understood that CTRL + ATL + DEL is doing the trick for my case because it does enter into the password field onto Windows 10 without entering any character even if the focus has been already acquired. I tested this option and it works well in my use case.

I did not used that feature option before because it does not mention "Password" in the ComboBox as seen in the picture here:
![image](https://user-images.githubusercontent.com/31821960/185636582-57a4be19-fe53-4926-8d25-cfa074c25b88.png)

I though it could be only a translation mistake for french, but I checked in the .ts file, the original text in english does not mention "Password" too.

That said, the code is explicit enough to know that the password is actually entered after the key combination:
![image](https://user-images.githubusercontent.com/31821960/185640462-c92f959a-286d-45af-843b-7004c1199fe1.png)

But for some reason (maybe text length?) the "Password" or at least "Pass" word has been omitted. It is not clear enough for the user (at least to me) what it actually does.

[...]

PR #1066, opened 2022-08-30
Fix issue #1052 and update french translation

fc-projects:
Improve readability of what is a key combination by putting it inside brackets and removing useless spaces.
Add the missing entry of Ctrl+Alt+Del without unlock.
Update the whole french translation.


limpkin:
thanks a lot :), it's really appreciated!

PR #1077, opened 2022-10-13
[BLE] Add option to correct serial number

deXol:
For devices with above 2000 serial number compare serial number with platform serial, if they are not equal display error message and option to set the correct serial number.
setSerialDemo


limpkin:
congrats!

PR #1125, opened 2023-01-20
add disable TLD check option

### LINKED ISSUE #1100: Add option to not check TLDs again existing list when entering domain extensions for the "multiple domains" functionality.
Currently, if you enter a TLD/extension that's not on the list included in the Moolticute code, it's silently ignored. For example adding ".com.be" for Amazon, which is not an officially designated TLD but a company that bought com.be and is selling subdomains. But similarly sometimes companies have internal TLDs, same for developers using something like .local. In my opinion it should first of all error and not silently throw away certain items. As discussed on IRC, including an option to accept unknown extensions would be nice. I would suggest not making this a general option, but rather adding it to the small dialog, for example a checkbox underneath the text box with something like "allow non-standard extensions".

Screenshot at 2022-12-04 04-29-45

rsrdesarrollo:
fix #1100 

It adds a new config option checkbox and skips TLD validation when setting is enabled.

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>946</width>
-    <height>1120</height>
+    <height>632</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -576,8 +576,8 @@ Hint: keep your mouse positioned over an option to get more details.</string>
             <rect>
              <x>0</x>
              <y>0</y>
-             <width>1014</width>
-             <height>1721</height>
+             <width>770</width>
+             <height>882</height>
             </rect>
            </property>
            <layout class="QVBoxLayout" name="verticalLayout_2">
@@ -3901,7 +3901,7 @@ Hint: keep your mouse positioned over an option to get more details.</string>
                <x>0</x>
                <y>0</y>
                <width>646</width>
-               <height>941</height>
+               <height>874</height>
               </rect>
              </property>
              <property name="autoFillBackground">
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>946</width>
-    <height>1120</height>
+    <height>632</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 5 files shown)


bertvandepoel:
I'm not very good at compiling Qt apps on my (relatively old) OS 😓 Would you mind including a screenshot? I'm hoping to help with the UX. While moolticute is very useable for IT people, it's proving to be really difficult for more regular users. I'm hoping to help improve that :)

src/MainWindow.ui
```diff
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>946</width>
-    <height>632</height>
```
deXol, line 10: Can you revert this please? It is an automatically updated value, when you open the UI file it is updating it based on your Qt window size.

src/MainWindow.ui
```diff
@@ -576,8 +576,8 @@ Hint: keep your mouse positioned over an option to get more details.</string>
             <rect>
              <x>0</x>
              <y>0</y>
-             <width>770</width>
-             <height>882</height>
+             <width>1014</width>
```
deXol, line 579: Same here as in line 10.

src/MainWindow.ui
```diff
@@ -3930,7 +3901,7 @@ Hint: keep your mouse positioned over an option to get more details.</string>
                <x>0</x>
                <y>0</y>
                <width>646</width>
-               <height>874</height>
+               <height>941</height>
```
deXol, line 3904: Same here as in line 10.

deXol (changes requested):
Hello,
Functionality looks good, but I built it and Settings page looks really strange:
kép
Moolticute Settings text moved to the bottom, text is missing for the second setting.
Original UI:
kép

rsrdesarrollo:
@deXol done!

bertvandepoel:
Does this PR exclusively add a setting, or does it also include some kind of warning or other indication when TLDs are listed that are not in the built-in list? I must admit I'm really not a fan of current behaviour where they are silently ignored (though I'm not sure that discussion is "in scope" for this PR)

bertvandepoel:
Awesome! :D Nevermind about what I said then!

deXol:
@rsrdesarrollo Window size constants are good now, but the Settings page is still off:
kép
Moolticute Settings text should be on the top and View daemon log should be a line above.

rsrdesarrollo:
@deXol now is really done.

rsrdesarrollo:
and now is really really done xD 

P.D: I hate qt-creator

deXol (approved):
nice, looks good now :)

limpkin:
thanks a lot @rsrdesarrollo !

PR #1131, opened 2023-01-22
Generate Exec in moolticute.desktop during build

VincentVanlaer:
To be merged after #1130, otherwise builds for the debian package will be broken (`PREFIX` will not be replaced).

Depending on how moolticute is installed or which distribution one uses, the moolticute executable might not be located in `/usr/bin`. For example, the nix package manager installes packages under `/nix/store`. This would also fix manual installs in `/usr/local`.


bobsaintcool:
This makes sense, thanks for you contribution

PR #1213, opened 2024-02-25
Add TOTP secret from QR code, fix #1044

### LINKED ISSUE #1044: Add TOTP secret from QR code
### Missing feature
Add TOTP via QR code, e.g. have a button to
1. Take a screenshot and read the QR code from it, or
2. Read QR code from a saved image

Possibly this would be less of an issue with proper TOTP browser integration (where the browser extension could automatically figure out that an image is a QR code for TOTP, or the user could click on the image directly)

### Justification
Some sites only show a QR code for TOTP, and don't have a way of showing the secret as text.

### Workarounds
Save the QR code, use an external QR decoder application to get the secret, and insert it in Moolticute.

deXol:
Integrated https://github.com/ftylitak/qzxing library for reading QR code.
Added wrapper for TOTP reading.

### Implementation
**TOTP QR format:** 
`otpauth://totp/<service>:<login>?PARAMETERS`
**Example**: 
`otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example`

### Implementation
Currently two ways of TOTP QR read is implemented:
1. **Create a screenshot of TOTP QR**
Screenshot is detected when user is in MMM and tries to find the credential provided in TOTP URI and add TOTP for credential, if service or login is not found after user confirmation add service/login and TOTP.
**In MMM screenshot is detected and creates non-exising credential:**
[Example](https://github.com/mooltipass/moolticute/assets/11043249/40ee97fd-7584-48ad-99d0-e55b02bf6fa5)
**Process clipboard data after entering MMM:**
[Example](https://github.com/mooltipass/moolticute/assets/11043249/7a825d33-0140-4fa1-a35f-0e7919479ffb)
Screenshot the TOTP QR image and enter MMM, after entering TOTP is detected from clipboard.
In TOTP QR service is gmail, this service is not available in DB, but it has a gmail.com service, so during scan QR Moolticute asks the user if they want to add it for this service.

2. **Scan QR for given credential**
For "Setup TOTP Credential" a QR option is added which will fill the available datas from TOTP QR image:
[Example](https://github.com/mooltipass/moolticute/assets/11043249/c14f7929-bdec-4896-bae5-190cdb41f998)

### TODOs
- ~Improve UI for TOTP reading
For testing I added a Scan QR button for reading TOTP QR and processing it.~
- ~Add solution to paste QR image from clipboard~
- Improve dialog texts during adding TOTP
- ~TOTP storage for credential with multiple domain fails. (not related to QR read),~ **fixed with #1222** 
- ~Test with other TOTP examples~
- ~Test with older unix build~ (Works on xenial, bionic)~

### Note
On **Mac** there is an issue with processing clipboard dataChanged signal (https://doc.qt.io/qt-6/qclipboard.html#dataChanged).
As a workaround I implemented a check in MMM, when Moolticute app is activated in MMM then I am checking the content of the clipboard. An image is only processed once.


src/CredentialModel.cpp
```diff
@@ -307,15 +307,25 @@ qint8 CredentialModel::getAvailableFavorite(qint8 newFav)
     }
 }
 
-QModelIndex CredentialModel::getServiceIndexByName(const QString &sServiceName, int column) const
+QModelIndex CredentialModel::getServiceIndex(const QString &sServiceName, Qt::MatchFlag flag, int column) const
```
limpkin, line 310: is the matchflag compatible with old qt versions? :)

deXol: Yes, Qt::MatchExactly value was already used from the enum earlier. (line 312)

limpkin:
that's a lot of files... could we use submodules instead, in case bug are detected in our dependencies?

PR #1214, opened 2024-02-27
Remove window size workaround on Unix

principis:
The Qt6 versions of Moolticute on KDE fails to launch due to the workaround in #972. I tried to reproduce the issue this workaround should fix on multiple builds and systems, but failed to so.

This PR removes that workaround.

### DIFF
--- a/src/AppGui.cpp
+++ b/src/AppGui.cpp
@@ -839,13 +839,6 @@ void AppGui::createMainWindow()
 
     win = new MainWindow(wsClient, dbMasterController);
 
-#ifdef Q_OS_UNIX
-    // Force resize and reset maximumWidth
-    // This fixes a crash on some Gnome+Wayland platforms
-    win->resize(0,0);
-    win->setMaximumWidth(win->width());
-#endif
-
     connect(win, &MainWindow::destroyed, [this](QObject *)
     {
         win = nullptr;


limpkin:
@principis can we make this fix for QT6 builds only?

principis:
I would just remove it. I tested the commit before the fix on 2 different Ubuntu versions (can't remember was a while ago) and on Fedora. I think the issue is fixed in Mutter itself.

limpkin:
right... but for users having the same old linux version this patch was made for..... that would break their install, right?

principis:
> right... but for users having the same old linux version this patch was made for..... that would break their install, right?

Yes, but that shouldn't be possible as the issue was only reported on Fedora. I also tested it on multiple systems, but maybe not enough...

Would it be good enough if I make a list of distros to test on and check them all?

Requirements for the bug:
- Gnome (Mutter window manager) on Wayland
- Enable User Themes extensions
- Apply a different theme
- Test the latest commit and 36d5f4d94062c12245bcbf866db68773f939b000

Distros:
- [ ] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] Ubuntu 23.10
- [ ] Fedora 38
- [ ] Fedora 39
- [ ] Fedora 40

Let me know which ones you would like me to add.

limpkin:
actually, given that you were the one who made this patch, I'll merge as is :)

PR #1234, opened 2024-07-28
lang: Update russian translation

nekromant:
This PR fully reworks Russian translation for all of the settings present in the app, fixes typos present, perhaps adds new ones.


limpkin:
oh wow, thanks so much for spending so much time doing this.... you're the reason that motivates us to keep going forward :)

nekromant:
@limpkin You're welcome. Please check the email, I've send a request to get all that's needed to get the actual firmware translated as well.

PR #1235, opened 2024-07-30
Synchronize all translations and update russian translation

nekromant:
It seems that some strings were missing, so I took the liberty of running lupdate to sync all of the translations with the current source (commit 1) and later translate the remaining strings. 

The only thing's missing right now is the translation of the device firmware itself.

### DIFF
--- a/lang/mc_br.ts
+++ b/lang/mc_br.ts
@@ -12,118 +12,118 @@
 <context>
     <name>AppGui</name>
     <message>
-        <location filename="../src/AppGui.cpp" line="130"/>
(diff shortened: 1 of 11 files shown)


limpkin:
thank you so much! did you receive my emails with the files you requested?

nekromant:
@limpkin It seems that there's a problem at my email server with sending and receiving emails from proton.me domains. I got the second email with the question somehow, but couldn't reply - the email won't go through. Looks like ISP blockage. I've sent you a backup email address via website form.

PR #1244, opened 2024-09-18
Add 'Community creation' warning

nekromant:
I have made a few mini bles for me and friends based on the available specs. As @limpkin suggested I used serial numbers >= 100000 for them. To avoid any frustration (in case any of the devices end up on ebay), I suggest treating any devices with serial number >=100000 as 'community creations'. For them: 

- Security challenge is disabled to avoid any irrelevant spam to the (unexpecting) original team,
- A warning is printed that the device is not affiliated with the original team. 

The translation files are updated yet again, I've added relevant Russian translation for the new messages and (hopefully) not too ugly German one. (The latter one needs some checking by some native speaker though, my knowledge of German is pretty bad).

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>946</width>
-    <height>1081</height>
+    <height>632</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -576,8 +576,8 @@ Hint: keep your mouse positioned over an option to get more details.</string>
             <rect>
              <x>0</x>
              <y>0</y>
-             <width>932</width>
-             <height>1832</height>
+             <width>770</width>
+             <height>882</height>
             </rect>
            </property>
            <layout class="QVBoxLayout" name="verticalLayout_2">
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -3955,8 +3955,8 @@ Hint: keep your mouse positioned over an option to get more details.</string>
               <rect>
                <x>0</x>
                <y>0</y>
-               <width>524</width>
-               <height>704</height>
+               <width>646</width>
+               <height>885</height>
               </rect>
              </property>
              <property name="autoFillBackground">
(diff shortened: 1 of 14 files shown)


src/MainWindow.ui
```diff
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>946</width>
-    <height>632</height>
+    <height>1081</height>
```
limpkin, line 10: i think that's a problem.... @deXol ?

src/MainWindow.ui
```diff
@@ -576,8 +576,8 @@ Hint: keep your mouse positioned over an option to get more details.</string>
             <rect>
              <x>0</x>
              <y>0</y>
-             <width>770</width>
-             <height>882</height>
+             <width>932</width>
+             <height>1832</height>
```
limpkin, line 580: tagging @deXol too

src/MainWindow.ui
```diff
@@ -3939,8 +3955,8 @@ Hint: keep your mouse positioned over an option to get more details.</string>
               <rect>
                <x>0</x>
                <y>0</y>
-               <width>646</width>
-               <height>885</height>
+               <width>524</width>
+               <height>704</height>
```
limpkin, line 3959: here too...

limpkin:
thanks a lot @nekromant !!! It's really appreciated.

src/MainWindow.ui
nekromant: I don't remember explicitly changing these sizes, must be qt creator's doing. Will double-check tomorrow.

nekromant:
Obligatory screenshot of how it looks like with a homebrew device: 

Screenshot_20240918_233838

src/MainWindow.ui
nekromant: Fixed

src/MainWindow.ui
nekromant: fixed

src/MainWindow.ui
nekromant: also fixed

limpkin:
@deXol could you review this one? :)

deXol (approved):

PR #1245, opened 2024-10-12
Set layout based on system's layout on first connection, fix #1197

### LINKED ISSUE #1197: Entering recorded data regardless of system layout
### Missing feature
https://github.com/mooltipass/minible/issues/413
### Justification
This update will help make the device easier to use for those who have more than one layout on their system.

This will also allow you to enter credentials with mixed layouts.

### Workarounds
Manual layout switching

deXol:
Add tracking for first device connection. (based on serial number)
When device is connected first time to the given system, set layouts according to system's layout.


src/utils/DeviceConnectionChecker.cpp
```diff
@@ -0,0 +1,46 @@
+#include "DeviceConnectionChecker.h"
+#include "AppGui.h"
+
+const QString DeviceConnectionChecker::CONNECTED_DEVICES_GROUP = "ConnectedDevices/";
```
limpkin, line 4: out of curiosity, we didn't have a similar service before? for example, in our logic to check the change number for a given user

limpkin (changes requested):

src/Settings/SettingsGuiHelper.cpp
```diff
@@ -261,6 +261,7 @@ void SettingsGuiHelper::updateParameters(const QJsonObject &data)
         val = data["value"].toBool();
     }
     m_settings->updateParam(m_settings->getParamId(param), val);
+    m_mw->checkSettingsChanged();
```
limpkin, line 264: out of curiosity, what's that for?

src/utils/DeviceConnectionChecker.cpp
deXol: We have DbExportsRegistry: https://github.com/mooltipass/moolticute/blob/master/src/DbExportsRegistry.cpp
It is embedded in DbBackup codes, so was not able to reuse it, it writes dbExportsRegistry.ini with the changeNumbers for given cardId.
I would be able to use that file, but I believe a new connectedDevices.ini much more clean.

src/Settings/SettingsGuiHelper.cpp
deXol: src/utils/KeyboardLayoutDetector.cpp:23 I am sending parameters directly for layouts, it updates these parameters in daemon and send back param_changed message, which calls this function and setting UI element to correct value (in this case the layout). However settingsChanged is not checked and due to the new layout Save/Cancel buttons are displayed without this call.

src/utils/KeyboardLayoutDetectorMac.cpp
```diff
@@ -7,55 +7,63 @@ KeyboardLayoutDetectorMac::KeyboardLayoutDetectorMac(QObject *parent)
...
-        {"com.apple.keylayout.Slovenian","Slovenia"},
-        {"com.apple.keylayout.Spanish","Spain"},
-        {"com.apple.keylayout.Spanish-ISO","Spain"},
-        {"com.apple.keylayout.Swedish","Sweden & Finland"},
-        {"com.apple.keylayout.Swedish-Pro","Sweden & Finland"},
-        {"com.apple.keylayout.Finnish","Sweden & Finland"},
-        {"com.apple.keylayout.SwissFrench","Swiss French"},
-        {"com.apple.keylayout.British","United Kingdom"},
-        {"com.apple.keylayout.British-PC","UK Extended"},
+        {"com.apple.keylayout.Dutch","Nertherlands (Mac)"},
```
deXol, line 42: typo on layout name fetched from device

limpkin:
sorry for the delay @deXol , very nice work!

PR #1252, opened 2025-07-08
Main window size

lcp5y3:
# Main window display

In french version of moolticute, some pane of the main window have widgets partially hide (see image below). So some action aren't accessible. To fix this issue, I've increase `MainWindow->maximumSize->width` value from `946` to `16777215` (the same than height).

After this, moolticute can be display with full screen and all widget are accessible.

moolticute

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -18,7 +18,7 @@
   </property>
   <property name="maximumSize">
    <size>
-    <width>946</width>
+    <width>16777215</width>
     <height>16777215</height>
    </size>
   </property>


limpkin:
@deXol do you remember why we weren't doing this before? I remember display issues too....

lcp5y3:
Any update ?

limpkin:
@deXol ?

deXol:
> @deXol ?

We just kept the default width at 946 value as a fix at line 9, because it worked on most systems and Qt UI editor changes that value all the time you open the editor.
I cannot see problem with changing maxWidth to 16777215 as for max height.

deXol (approved):

Issue #1254, opened 2026-05-05
build for Apple silicon (arm) macOS


simtomas:
Hi,

it looks like macOS 27 will no longer have Rosetta and will not run Intel binaries.
Could you please create a arm build for macOS?

hyperbotsx:
Native Apple Silicon builds are now working and hardware-validated: PR #1255 (build pipeline, by @poitee) plus https://github.com/poitee/moolticute/pull/1 (hardening + a critical arm64 daemon crash fix found in device testing) together produce a fully native DMG via GitHub Actions that passes the complete USB feature matrix on a Mooltipass Mini BLE under macOS 26.5.2 — no Rosetta involved. BLE-over-Bluetooth testing to follow. Given macOS 26.4+ already warns on every Rosetta app launch and general Rosetta support ends with macOS 28, it would be great to see these land in an official release.

Issue #929, opened 2021-11-02
moolticute Apple M1 support


WebmasterTD:
### [Question] Apple M1 support
Is there a plan to add support for Apple M1 native builds?
The transition  from QT 5.x to version 6.2+ on the roadmap?

limpkin:
@raoulh will confirm, but we're indeed planning to transition to qt 6.x once the open source version is released (if it hasn't already).
Regarding M1 support... @raoulh will provide some feedback :)
tagging @deXol

raoulh:
Qt 6.2 is out and it supports Apple M1. The port can be done then

TLeM4:
I have an Apple M1 if it's necessary to test the app.
This is the only app i need on a daily basis that need Apple Rosetta2 emulation.

ananthb:
I can volunteer testing effort on my hardware as well.
I can also contribute code towards this if someone can nudge me in the right direction.
I have not worked with Qt before but I know my way around C++.

limpkin:
@raoulh could you give it a look now that we're on qt6?

PR #1255, opened 2026-06-17
Add native Apple Silicon macOS build support

poitee:
## Summary
- Add `scripts/macos/build-local.sh` for native Intel and Apple Silicon builds using Homebrew Qt
- Update macOS CI scripts to detect architecture, use Homebrew Qt on arm64, and compile `mc-agent`/`mc-cli` from source instead of bundling Intel-only binaries
- Add GitHub Actions workflow for macOS builds on `macos-14`
- Document the build process in `documentation/macos_apple_silicon.md`

## Motivation
Official release DMGs are `x86_64` only. On Apple Silicon Macs, Moolticute currently requires Rosetta ([#1254](https://github.com/mooltipass/moolticute/issues/1254), [#929](https://github.com/mooltipass/moolticute/issues/929)).

## Test plan
- [x] `./scripts/macos/build-local.sh` on Apple Silicon (arm64) — produces native binary
- [x] `./scripts/macos/build-local.sh --package` bundles arm64 `mc-agent` and `mc-cli`
- [x] `./build/tests/tests` — 23 tests passed
- [ ] Maintainer verification with Mini BLE over USB on Apple Silicon hardware

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/macos-build.yml
+++ b/.github/workflows/macos-build.yml
@@ -1,0 +1,31 @@
+name: macOS build
+
+on:
+  push:
+    branches: [master]
+    tags: ['v*']
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install dependencies
+        run: brew install qt go
+
+      - name: Build
+        run: ./scripts/macos/build-local.sh
+
+      - name: Verify architecture
+        run: |
+          file build/Moolticute.app/Contents/MacOS/moolticute
+          lipo -info build/Moolticute.app/Contents/MacOS/moolticute
+
+      - name: Run tests
+        working-directory: build
+        run: ./tests/tests
--- a/.github/workflows/macos-build.yml
+++ b/.github/workflows/macos-build.yml
@@ -2,7 +2,7 @@ name: macOS build
 
 on:
   push:
-    branches: [master]
+    branches: [master, feature/apple-silicon-macos]
     tags: ['v*']
   pull_request:
     branches: [master]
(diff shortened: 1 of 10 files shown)


poitee:
I testing via real hardware, it compiles and seems to work. Hope this helps.

hyperbotsx:
I've hardware-validated this PR's approach on an Apple Silicon Mac (macOS 26.5.2) with a Mooltipass Mini BLE, and contributed a set of hardening commits plus one critical fix on top of it: https://github.com/poitee/moolticute/pull/1

Headline results:

- The pipeline approach works — GitHub Actions `macos-14` (arm64) runners build a fully native bundle (GUI, daemon, mc-agent, mc-cli all `arm64`), 63/63 unit tests pass in CI.
- **Testing on real hardware caught a showstopper that affects every native arm64 build of master, independent of this PR:** the daemon crash-loops (`EXC_BREAKPOINT` in `FilesCache::setCardCPZ`) the moment a card CPZ is read — an undefined 32-bit shift in the SimpleCrypt key derivation that x86 tolerated and arm64 compiles into a trap. Fixed with byte-identical key derivation (details + binary-level evidence in the linked PR). After the fix, the full USB feature matrix passes on my device: detection, unlock, credential read/write, browser extension, mc-cli, notes, settings, replug recovery.
- BLE testing is still pending on my side (device battery died mid-session) — will follow up with results.
- Two more findings worth knowing: the auto-updater offers the Intel v1.04.0 DMG to native builds (accepting the "update" would reinstall the Rosetta dependency), and upgrading over an Intel install can leave a stale shared-memory daemon record that blocks the new daemon (documented a workaround in the docs).

Thanks @poitee for the excellent base — hopefully the extra validation helps get this merged.

limpkin:
@raoulh could you give it a look? i'm not proficient enough in this...

